PR #12, opened 2016-01-28
Implemented Fallback Logic for fragment

vigneshshanmugam:


w0rm:
Can you add a test for the correct response code of tailor, when the primary fails, and fallback is fetched?

vigneshshanmugam:
Added and rebased the commits

PR #36, opened 2016-06-27
Increase listeners for concurrent handlers on stream

vigneshshanmugam:
`Warning: Possible EventEmitter memory leak detected. 11 data listeners added. Use emitter.setMaxListeners() to increase limit`.

Addresses this error, by keeping the listeners in control. This will make sure we get error message if an actual leak occurred.


grassator:
👍

mo-gr:
👍

PR #41, opened 2016-07-12
add missing break in example

timsim00:
Small change but when break is missing causes issues when trying to convert example to pull css from files.


lmineiro:
@w0rm @vigneshshanmugam ?

w0rm:
@timsim00 thanks!

PR #44, opened 2016-07-19
Partial Templates and AST

vigneshshanmugam:
PR enables these ideas. 

- [x] Support Partial Template
- [x] AST Parsing/ HTML compatible
- [x] Slot Based Transforms
- [x] Allow custom parent/child template caching


vigneshshanmugam:
Just to clarify some things about partial templates. To make it more compatible with the spec. If you want fragment or slot to be in the head, then you need to define them by this way
```html
<script type="slot" name="head"></script>
<script type="fragment" public src="https://mosaic01.ztat.net/base-assets/fragment.html">
</script>
```

if you don't need it in the head, you can also define them by this way
```html
 <body>
      <slot name="before-body"></slot>
      <fragment src="http://localhost:8088" primary fallback-src="http://localhost:8081"></fragment>
</body>
```

lib/serializer.js
```diff
@@ -0,0 +1,133 @@
...
+        if (attribs && attribs.type) {
+            return node.name === 'script' && this.handleTags.indexOf(attribs.type) !== -1;
+        }
+        return false;
+    }
+
+    _isLastChildOfBody(node) {
+        const parentNode = node.parent;
+        if (parentNode.name === 'body') {
+            if (Object.is(node, parentNode.lastChild)) {
```
grassator: Is there a reason to use Object.is and not tripl

grassator: Triple equals

lib/serializer.js
```diff
@@ -0,0 +1,133 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    _serializeSpecial(node) {
+        this.pushBuffer();
+        let fragmentObj;
+        if (this.handleTags.indexOf(node.name) !== -1) {
```
grassator: Not a necessary change but t probably would be more readable with "includes" from ES6

lib/serializer.js
vigneshshanmugam: No both should do exactly the same.

lib/transform.js
```diff
@@ -0,0 +1,61 @@
...
+        };
+        const serializer = new CustomSerializer(rootNodes, serializerOptions);
+        return serializer.serialize();
+    }
+
+    _groupSlots(root) {
+        const slotMap = new Map();
+        slotMap.set('default', []);
+        const nodes = adapter.getChildNodes(root);
+        nodes.forEach((node) => {
```
grassator: We should either explicitly tell that we only support slots on the top level or traverse all the nodes.

package.json
```diff
@@ -5,6 +5,7 @@
   "keywords": [
```
grassator: Since this is a backwards incompatible change we need to bump the major version and probably start doing proper releases and release notes.

lib/transform.js
vigneshshanmugam: Agreed.

lib/serializer.js
vigneshshanmugam: not available in 4.4

package.json
vigneshshanmugam: agreed.

example/fragment.js
```diff
@@ -42,5 +42,6 @@ module.exports = (fragmentName, fragmentUrl) => (request, response) => {
                     Fragment ${fragmentName}
                 </div>
             `);
+            break;
```
mo-gr: break at the end of default? why?

lib/serializer.js
```diff
@@ -0,0 +1,133 @@
...
+
+    _isSlotNode(node) {
+        if (node.name === 'slot') {
+            return true;
+        }
+        const attribs = node.attribs;
+        if (!attribs) {
+            return false;
+        }
+        return node.name === 'script' && attribs.type === 'slot';
```
mo-gr: not sure, this improves readability a lot. but this could also be written as:
`return node.name === 'script' && attribs && attribs.type === 'slot'`

package.json
```diff
@@ -22,7 +23,7 @@
     "test"
   ],
   "engines": {
-    "node": ">5.0.0"
+    "node": ">4.4.x"
```
mo-gr: wow. we went backwards in the required node engine? not bad...

mo-gr:
👍

vigneshshanmugam:
addressed the concerns @mo-gr

PR #47, opened 2016-07-26
Remove slots attributes when rendering

vigneshshanmugam:


lib/transform.js
```diff
@@ -44,6 +44,7 @@ module.exports = class Transform {
                     slotMap.set(attribs.slot, [node]);
                 }
                 this._pushText(node.next, slotMap.get(attribs.slot));
+                delete attribs.slot;
```
grassator: I'm not sure if copying all attributes besides slot would be faster. Would be interesting to see if attribs is a HashMap or V8 transforms it into a class...

vigneshshanmugam: `attribs` is in FastMode before and after the deletion its changed to slowDictionary mode in v8.  We can do the same hack which bluebird does to make sure the object is in fast mode even after deletion. 

```js
function toFastProperties(obj) {
    /*jshint -W027*/
    function f() {}
    f.prototype = obj;
    ASSERT("%HasFastProperties", true, obj);
    return f;
    eval(obj);
}
```

vigneshshanmugam: I tested it in our code and object is always in fast mode even after calling delete on it.

Issue #48, opened 2016-08-03
handleTag functionality is broken for tags which has childNodes


vigneshshanmugam:

vigneshshanmugam:
#50 Will address this issue.

PR #50, opened 2016-08-03
add support for flattening nested fragments

vigneshshanmugam:
This PR addresses two things. 
+ fixes the issue with handle tags #48 
+ Add support for the nested fragments


grassator:
Seems the build is broken...

PR #51, opened 2016-08-04
enable rendering with presentable erros

vigneshshanmugam:


lib/request-handler.js
```diff
@@ -124,7 +124,12 @@ module.exports = function processRequest (options, request, response) {
             if (shouldWriteHead) {
                 shouldWriteHead = false;
                 response.writeHead(500, responseHeaders);
-                response.end();
+                // To render with custom error template
+                if (err.presentable) {
```
grassator: I wouldn't do 2 fields because then you can get into a state where you say it's presentable but don't actually have anything to show. Just checking if something is there should be enough

grassator:
👍

PR #52, opened 2016-08-17
changes to remove the binding

vigneshshanmugam:


lib/request-handler.js
```diff
@@ -60,13 +60,12 @@ module.exports = function processRequest (options, request, response) {
                 );
 
                 FRAGMENT_EVENTS.forEach((eventName) => {
-                    fragment.on(eventName, (function () {
```
w0rm: we can try to use `(...args) => {}` instead of function here

vigneshshanmugam: changed it.

w0rm:
👍

PR #54, opened 2016-08-26
introduce unnamed slot behaviour

vigneshshanmugam:
This will allow us to insert contents anywhere in the parent template without ugly slot name magic in the child templates.

### DIFF
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -83,8 +83,9 @@ module.exports = class CustomSerializer extends Serializer {
             if (slots) {
                 slots.forEach(this._serializeNode);
             }
+        } else if (this.defaultSlotsInserted) {
+            console.warn('Encountered duplicate Unnamed slots in the template - Skipping the node');
         } else {
-            // Unnamed slot behaviour
             const defaultSlots = this.slotMap.get('default');
             this.defaultSlotsInserted = true;
             if (defaultSlots) {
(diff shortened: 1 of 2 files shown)


lib/serializer.js
```diff
@@ -77,11 +78,18 @@ module.exports = class CustomSerializer extends Serializer {
...
-        if (this.slotMap.has(slotName)) {
+        if (slotName && this.slotMap.has(slotName)) {
             const slots = this.slotMap.get(slotName);
             if (slots) {
                 slots.forEach(this._serializeNode);
             }
+        } else {
+            // Unnamed slot behaviour
+            const defaultSlots = this.slotMap.get('default');
+            this.defaultSlotsInserted = true;
```
grassator: We should check what does the spec say about duplicate default slots. Maybe we should also warn on that.

lib/serializer.js
```diff
@@ -77,11 +78,18 @@ module.exports = class CustomSerializer extends Serializer {
...
         const slotName = node.attribs.name;
-        if (this.slotMap.has(slotName)) {
+        if (slotName && this.slotMap.has(slotName)) {
             const slots = this.slotMap.get(slotName);
             if (slots) {
                 slots.forEach(this._serializeNode);
             }
+        } else {
+            // Unnamed slot behaviour
+            const defaultSlots = this.slotMap.get('default');
```
grassator: Where do you "set" that unnamed slot as a "default"?

vigneshshanmugam: [Here](https://github.com/zalando/tailor/blob/master/lib/transform.js#L33) i am grouping them together to default slots if they don't have a slot name

lib/serializer.js
vigneshshanmugam: checked the spec, Sadly nowhere it was mentioned

grassator:
👍

PR #58, opened 2016-08-31
update loadCss to fix firefox crash issue on 38

vigneshshanmugam:
- This fixes the regression issue on FF 
- Issue here - https://bugzilla.mozilla.org/show_bug.cgi?id=1203973


vigneshshanmugam:
check the loadCss code here - https://github.com/filamentgroup/loadCSS/blob/master/src/loadCSS.js

grassator:
👍

PR #59, opened 2016-09-02
capture metrics for fragment initialization

vigneshshanmugam:
+ Will allow us to capture the initialisation cost for each fragments on the page.


grassator:
👍

Issue #55, opened 2016-08-26
Example doesn't work


akeelnazir:
Exception:
--------------
~/code/tailor/tailor-src$ npm run example

> node-tailor@1.0.1 example /Users/akeelnazir/code/tailor/tailor-src
> npm run prepublish && node --harmony example/tailor

> node-tailor@1.0.1 prepublish /Users/akeelnazir/code/tailor/tailor-src
> uglifyjs src/pipe.js --mangle --compress --comments -o src/pipe.min.js

Tailor started at port 8080
Fragment1 started at port 8081
Fragment2 started at port 8082
Fragment3 started at port 8083
fs.js:634
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/Users/akeelnazir/code/tailor/tailor-src/example/templates/.html'
    at Error (native)
    at Object.fs.openSync (fs.js:634:18)
    at Object.fs.readFileSync (fs.js:502:33)
    at /Users/akeelnazir/code/tailor/tailor-src/lib/fetch-template.js:11:31
    at Tailor.processRequest (/Users/akeelnazir/code/tailor/tailor-src/lib/request-handler.js:24:29)
    at emitTwo (events.js:106:13)
    at Server.emit (events.js:191:7)
    at HTTPParser.parserOnIncoming [as onIncoming] (_http_server.js:543:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:105:23)

filename: npm-debug.log
--------------
0 info it worked if it ends with ok
1 verbose cli [ '/Users/akeelnazir/n/bin/node',
1 verbose cli   '/Users/akeelnazir/n/bin/npm',
1 verbose cli   'run',
1 verbose cli   'example' ]
2 info using npm@3.9.3
3 info using node@v6.2.1
4 verbose run-script [ 'preexample', 'example', 'postexample' ]
5 info lifecycle node-tailor@1.0.1~preexample: node-tailor@1.0.1
6 silly lifecycle node-tailor@1.0.1~preexample: no script for preexample, continuing
7 info lifecycle node-tailor@1.0.1~example: node-tailor@1.0.1
8 verbose lifecycle node-tailor@1.0.1~example: unsafe-perm in lifecycle true
[...]

PR #66, opened 2016-09-28
adding helper function to inject a template for root path

Amberlamps:
I haven't not yet worked with tailor more than the example, but as mentioned in #55 it can be confusing that the example is not listening on the root path. I do not know whether this feature is a valid use-case for production ready tailor use, but for the example it helps injecting a template that is used for the "/" path.
It also does not break any legacy implementations as "inject-root-template.js" wraps around "fetch-template.js".

### DIFF
--- a/example/tailor.js
+++ b/example/tailor.js
@@ -4,10 +4,14 @@ const path = require('path');
 const Tailor = require('../index');
 const fetchTemplateFs = require('../lib/fetch-template');
 const serveFragment = require('./fragment');
+const injectRootTemplate = require('../lib/inject-root-template');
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
@Amberlamps First of all, Thanks for the PR. 

I am planning to port the Zalando version of the tailor and merge it with the current repo. Doing that will change the logic of fetch template and introduces a lot of changes in the way templates are handled. So right now, Its not the best time to merge this PR.

I hope you are okay with closing this.

Amberlamps:
Alright. Thanks for your input!

Issue #70, opened 2016-10-13
default template fetching reports template not found even when it is a different issue


aryszka:
the error handling could use the err.code property to check if it is not found, or there is a different reason. In case of not found, the hosting app can choose to return 404, and e.g. 500 otherwise. Even better, it would be probably nicer if the promise was simply rejected with an error, wrapping the original error.

aryszka:
probably the advice to #69 applies here, too?

vigneshshanmugam:
landed in https://github.com/zalando/tailor/pull/71

PR #71, opened 2016-10-13
Builtin fetch template: async and handle not found

### LINKED ISSUE #69: improve the default behavior of template fetching
The template fetching typically is customized by the hosting application, but it would be better, even if for quick prototyping only, if it was asynchronous. (https://github.com/zalando/tailor/blob/master/lib/fetch-template.js#L12)

aryszka:
Closes #69 and #70

### DIFF
--- a/lib/fetch-template.js
+++ b/lib/fetch-template.js
@@ -4,21 +4,28 @@ const path = require('path');
 const url = require('url');
 const fs = require('fs');
 
+const TEMPLATE_ERROR = 0;
 const TEMPLATE_NOT_FOUND = 1;
 
-const fsError = (err) => {
-    if (err.code === 'ENOENT') {
-        return {code: TEMPLATE_NOT_FOUND, original: err};
-    }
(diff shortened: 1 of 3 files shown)


lib/fetch-template.js
```diff
@@ -4,25 +4,58 @@ const path = require('path');
...
+                    return parseTemplate(baseTemplate);
                 }
-            }
-            return Promise.resolve(parseTemplate(baseTemplate, pageTemplate));
-        } catch (e) {
-            console.error(e);
-            return Promise.resolve(new Buffer('Template not found'));
-        }
+
+                pageTemplate = baseTemplate;
```
vigneshshanmugam: @aryszka you forgot to initialise it :)

vigneshshanmugam: I fixed it and added logging on error :)

aryszka: yup, thx :)

PR #74, opened 2016-11-03
allow AMD loader to be inlined (closes #73)

### LINKED ISSUE #73: Inline the AMD to avoid the sync blocking script call
- Use https://github.com/grassator/iamdee

vigneshshanmugam:
+ inlining small AMD loader would improve our first paint time. 
+ backwards compatible with specifying AMD loader url.

### DIFF
--- a/example/tailor.js
+++ b/example/tailor.js
@@ -1,14 +1,11 @@
 'use strict';
 const http = require('http');
 const path = require('path');
-const fs = require('fs');
 const Tailor = require('../index');
 const fetchTemplateFs = require('../lib/fetch-template');
 const serveFragment = require('./fragment');
 const baseTemplateFn = () => 'base-template';
-// You can also pass AMD URL
-// const AMD_LOADER = 'https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.22/require.min.js';
-const AMD_LOADER = fs.readFileSync(require.resolve('iamdee'), 'utf-8');
+const AMD_LOADER = 'file://' + require.resolve('iamdee');
 const tailor = new Tailor({
     amdLoaderUrl: AMD_LOADER,
     fetchTemplate: fetchTemplateFs(path.join(__dirname, 'templates'), baseTemplateFn)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,20 +1,28 @@
...
-const AMD_LOADER_URL = 'https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.22/require.min.js';
 
 module.exports = class Tailor extends EventEmitter {
 
     constructor (options) {
         super();
-        const amdLoaderUrl = options.amdLoaderUrl || AMD_LOADER_URL;
+        const pipeChunk = (amdLoaderUrl, pipeInstanceName) => {
+            let definition;
+            if (amdLoaderUrl.startsWith('http')) {
```
grassator: This is not a good way. It's possible to have `/` or `//` urls when not serving from CDN

example/tailor.js
```diff
@@ -1,12 +1,16 @@
...
 const path = require('path');
+const fs = require('fs');
 const Tailor = require('../index');
 const fetchTemplateFs = require('../lib/fetch-template');
 const serveFragment = require('./fragment');
 const baseTemplateFn = () => 'base-template';
-
+// You can also pass AMD URL
+// const AMD_LOADER = 'https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.22/require.min.js';
+const AMD_LOADER = fs.readFileSync(require.resolve('iamdee'), 'utf-8');
```
grassator: To be honest I would accept `file://` urls and still use require.js by default since it's more battle-proven

example/tailor.js
```diff
@@ -5,8 +5,9 @@ const Tailor = require('../index');
 const fetchTemplateFs = require('../lib/fetch-template');
 const serveFragment = require('./fragment');
 const baseTemplateFn = () => 'base-template';
-
+const AMD_LOADER = 'file://' + require.resolve('iamdee');
```
grassator: If someone was relying on a full featured require.js this is a breaking change, I would keep it as a default for now.

vigneshshanmugam: but its an example...

PR #75, opened 2016-11-09
Support gzip in tailor

ytanruengsri:
Support gzip in tailor

### DIFF
--- a/lib/request-fragment.js
+++ b/lib/request-fragment.js
@@ -4,6 +4,8 @@ const http = require('http');
 const https = require('https');
 const url = require('url');
 const filterHeaders = require('./filter-headers');
+const zlib = require('zlib');
+const unzip = zlib.createGunzip();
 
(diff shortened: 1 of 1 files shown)


lmineiro:
seems to awfully simple that it makes me wonder :)

ytanruengsri:
@lmineiro need a support to test this PR

grassator:
@ytanruengsri One of the things missing here is that Tailor should actually send "Accept-Encoding" to fragments.

You can do a couple of things here. For the automated test you can look into how the fragments are tested (we emulate streams from what I remember), so you can just create your own stream and pipe it through gzip and then test the output.

Another thing is to try to start tailor locally and point one of the fragments to any website that uses gzip or, better, start another node.js service that will do that.

ytanruengsri:
see https://github.com/zalando/tailor/pull/76

PR #76, opened 2016-11-09
Support gzip in tailor 

ytanruengsri:
Support gzip in tailor


lib/filter-headers.js
```diff
@@ -5,7 +5,7 @@ module.exports = function filterHeaders (attributes, headers) {
     if (attributes.public) {
         return newHeaders;
     };
-    ['accept-language', 'referer', 'user-agent'].forEach((key) => {
+    ['accept-language', 'accept-encoding', 'referer', 'user-agent'].forEach((key) => {
```
grassator: We shouldn't proxy the browser, because we don't care for the fragments if the browser supports gzip.

.gitignore
```diff
@@ -5,3 +5,4 @@ coverage
 src/pipe.min.js
 jsconfig.json
 .vscode
+.idea
```
grassator: My personal opinion is that should live in your personal config (the same actually goes for .vscode, but that somehow slipped through)

tests/request-fragment.js
```diff
@@ -41,6 +42,34 @@ describe('requestFragment', () => {
...
+            .reply(200, () => {
+                const text = 'HTTP+GZIP';
+                const buf = new Buffer(text, 'utf-8');
+                return zlib.gzipSync(buf);
+            });
+
+        requestFragment('https://fragment/', fragmentAttrb, {
+            headers: {}
+        }).then((response) => {
+            let data = '';
```
grassator: I know this just a test but this is a wrong way to assemble chunks. Here's the right one:
https://github.com/zalando-incubator/perron/pull/6

vigneshshanmugam:
This shouldn't be in tailor.. This can be accomplished with some other middleware..

lmineiro:
@vigneshshanmugam you mean, since it's a library, whatever is using it should wrap gzip around it? What exactly does this mean for early flush and Smart Pipe(™)?

vigneshshanmugam:
Let me iterate why its a bad idea to do in fragment level.. 
+ gzip doesn't come for free. It requires significant CPU usage and doing it in fragment level would simply increase CPU proportional to number of fragments on the template. 
+ compression ratio would be pretty less by doing in fragments vs doing it on overall response. 

by piping it through a middleware, Both would be solved and we probably wont need this inside tailor itself.

vigneshshanmugam:
@lmineiro Was composing the reply before seeing your comment. Yup exactly, tailor would be wrapped around by gzip/compress middleware.

grassator:
@vigneshshanmugam For me there is always a fine line between composability and just doing the right thing by default, but you are probably right.

tests/request-fragment.js
```diff
@@ -30,17 +32,47 @@ describe('requestFragment', () => {
...
     });
 
+    it('Should request fragment using https protocol and gzip compression', (done) => {
+        nock('https://fragment')
+            .defaultReplyHeaders({
+                'content-encoding': 'gzip',
+            })
+            .get('/')
+            .reply(200, () => {
+                const text = 'HTTP+GZIP';
```
addi90: Small typo 'HTTPS+GZIP'

tests/request-fragment.js
```diff
@@ -30,17 +32,47 @@ describe('requestFragment', () => {
...
+        requestFragment('https://fragment/', fragmentAttrb, {
+            headers: {}
+        }).then((response) => {
+            const chunks = [];
+            response.on('data', (chunk) => {
+                chunks.push(chunk);
+            });
+            response.on('end', () => {
+                const data = Buffer.concat(chunks).toString('utf8');
+                assert.equal(data, 'HTTP+GZIP');
```
addi90: Small typo 'HTTPS+GZIP'

tests/request-fragment.js
```diff
@@ -30,17 +32,47 @@ describe('requestFragment', () => {
     it('Should request fragment using https protocol', (done) => {
         nock('https://fragment').get('/').reply(200, 'HTTPS');
         requestFragment('https://fragment/', fragmentAttrb, {headers: {}}).then((response) => {
-            let data = '';
+            const chunks = [];
             response.on('data', (chunk) => {
-                data += chunk.toString();
+                chunks.push(chunk);
```
addi90: Is it fine to not convert the `chunk` to string anymore?

vigneshshanmugam:
@ytanruengsri Many thanks for the PR first of all :)..

But if you agree to the above points. Could you please close it.. Thanks

ytanruengsri:
I agree with the middleware approach

vigneshshanmugam:
@ytanruengsri Since you have already worked on this implementation, I would like to proceed with your PR and modify some changes in that directly.. Can you give me access to your repo or I can clone it and give a PR to yours? 

Let me know what you think

PR #78, opened 2016-11-11
Support fragment level compression  

vigneshshanmugam:
+ picking @ytanruengsri commits  from #76 and did some modifications

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -4,6 +4,7 @@ const ContentLengthStream = require('./streams/content-length-stream');
 const EventEmitter = require('events').EventEmitter;
 const PassThrough = require('stream').PassThrough;
 const parseLinkHeader = require('parse-link-header');
+const zlib = require('zlib');
 
 const getAttributes = (tag, context) => {
     const attributes = Object.assign({}, tag.attributes);
@@ -81,6 +82,13 @@ module.exports = class Fragment extends EventEmitter {
 
         this.insertStart();
 
+        let responseStream = response;
+        const contentEncoding = response.headers['content-encoding'];
+        if (contentEncoding &&
+            (contentEncoding === 'gzip' || contentEncoding === 'deflate')) {
+            responseStream = response.pipe(zlib.createUnzip());
+        }
+
         const contentLengthStream = new ContentLengthStream((contentLength) => {
             if (!isFallback) {
                 this.emit('end', contentLength);
@@ -92,12 +100,12 @@ module.exports = class Fragment extends EventEmitter {
             this.stream.end();
         });
 
-        response.on('error', (err) => {
+        responseStream.on('error', (err) => {
             this.emit('warn', err);
             contentLengthStream.end();
         });
 
-        response
+        responseStream
             .pipe(contentLengthStream)
             .pipe(this.stream, {end: false});
     }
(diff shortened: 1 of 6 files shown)


ytanruengsri:
lgtm

lib/fragment.js
```diff
@@ -81,6 +82,13 @@ module.exports = class Fragment extends EventEmitter {
 
         this.insertStart();
 
+        let responseStream = response;
+        const contentEncoding = response.headers['content-encoding'];
+        if (contentEncoding &&
+            (contentEncoding === 'gzip' || contentEncoding === 'deflate')) {
+            responseStream = response.pipe(zlib.createUnzip());
```
tadjik1: You've added one more stream but haven't added one more 'error' handler for it, it can lead to error that will be thrown.

vigneshshanmugam: Thanks for the review.. Will fix that and update the PR.

grassator:
👍

lib/request-fragment.js
```diff
@@ -4,16 +4,21 @@ const http = require('http');
...
+};
 
 module.exports = function requestFragment (fragmentUrl, fragmentAttributes, request) {
     return new Promise((resolve, reject) => {
         const parsedUrl = url.parse(fragmentUrl);
         const options = Object.assign({
-            headers: filterHeaders(fragmentAttributes, request.headers),
+            headers: Object.assign(
+                filterHeaders(fragmentAttributes, request.headers),
+                defaultHeaders
```
mo-gr: This overwrites possibly filtered headers with the default headers. I understand, that this is what you want, but it breaks my expectation of "defaultHeaders". I think it should be renamed to something like "requiredHeaders" since they are always there no matter what filterHeaders sets.

vigneshshanmugam: Cool.. will rename it :)

tests/tailor.js
```diff
@@ -687,4 +691,68 @@ describe('Tailor', () => {
...
+    });
+
+    it('should close the streams properly during unzping error', (done) => {
+        nock('https://fragment')
+            .defaultReplyHeaders({
+                'content-encoding': 'gzip',
+            })
+            .get('/2')
+            .reply(200, ()=> {
+                return new Error('GZIP Error');
```
tadjik1: Is it correct handling of gzip error?
Seems like error on request object, not on gzip stream.
You can just pass normal string and set header to 'gzip', in that case unzip will fire an error 'incorrect header check'.

see an example:

```
describe('zlib stream error', () => {
  it('should handle error', done => {
    const sourceStream = new PassThrough();
    const destStream = new PassThrough();
    const unzipStream = zlib.createUnzip();

    sourceStream
      .on('error', () => console.log('source stream error'))
      .pipe(unzipStream)
      .on('error', err => {
        assert.equal(err.message, 'incorrect header check');
        done();
      })
      .pipe(destStream);

    sourceStream.end(Buffer.from('uncompressed content'));
  });
});

```

vigneshshanmugam: No its the error on response stream. Check the error it produces..
```js
{ Error: incorrect header check
    at Zlib._handle.onerror (zlib.js:370:17) errno: -3, code: 'Z_DATA_ERROR' }
```

tadjik1: oh, really, you totally right. interesting how nock provides error object as a response..

vigneshshanmugam:
Tests failing due to saucelabs token.. Merging it anyway..

PR #80, opened 2016-11-22
Redirect to index template instead of showing error

dazld:
as title


dazld:
closing as examples changed.

vigneshshanmugam:
Cool.. may be we should redirect all the images to / instead of `index`?

dazld:
could do, the only thing missing is some way to map `/` to `/index`. I kinda liked the redirect though, as it made it super clear what was happening.

PR #81, opened 2016-11-22
Implementation of Service Worker

vigneshshanmugam:
This is not the final implementation.. 

* [ ] Need to come up with API for Service Worker. 
* [ ] `sw.js` should be served from the current directory. 
* [ ] what if one of the fragment is down/timed out?


vigneshshanmugam:
closing this in favor of #84 .. Track the progress there.

PR #84, opened 2016-11-24
[WIP] Add service worker support

boopathi:


vigneshshanmugam:
* [ ] come up with API for Service Worker. 
* [ ] take fragment timeouts/async in to account? 
* [ ] create example and document the API.

vigneshshanmugam:
ping @mo-gr @grassator

lib/service-worker.js
```diff
@@ -0,0 +1,103 @@
...
+        response.end(`
+            var ${cssVarName} = [${css.map(c => JSON.stringify(c))}];
+            var ${jsVarName} = [${js.map(c => JSON.stringify(c))}];
+            ${options.serviceWorker.contents}
+        `);
+    });
+}
+
+function isObject(o) {
+    return !Array.isArray(o) && typeof o === 'object' && o !== null;
```
vigneshshanmugam: why not our good old friend?
```js
 Object.prototype.toString.call({})
```

lib/request-handler.js
```diff
@@ -138,16 +161,14 @@ module.exports = function processRequest (options, request, response) {
...
-                        resultStream.write(item);
-                    });
-                    resultStream.end();
-                } else {
-                    resultStream.end(parsedTemplate);
-                }
+            .then(_parsedTemplate => {
+                const parsedTemplate = Array.isArray(_parsedTemplate)
+                    ? _parsedTemplate
+                    : [_parsedTemplate];
```
dazld: `[].concat(_parsedTemplate)` instead of ternary ?

lib/service-worker.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function respondNotEnabled(response) {
```
dazld: would move this function definition to before it gets used. relying on hoisting can confuse.

Issue #83, opened 2016-11-23
Be html complaint for script tags passed via handledTags


vigneshshanmugam:
When the script `type` is set to `random-str` other than `text/javascript`.. the child contents should not be rendered. 

Happens only when the `type` is passed on the `handledTags` array.

PR #86, opened 2016-11-24
html compatible for script tags

vigneshshanmugam:
+ fix(#83)

### DIFF
--- a/lib/parse-template.js
+++ b/lib/parse-template.js
@@ -9,8 +9,7 @@ const Transform = require('./transform');
  * @returns {Promise} Promise that resolves to serialized array consisits of buffer and fragment objects
  */
 module.exports = (handledTags, insertBeforePipeTags) =>
-    (baseTemplate, childTemplate) => new Promise((resolve) => {
-        const transform = new Transform(handledTags, insertBeforePipeTags);
-        const serializedList = transform.applyTransforms(baseTemplate, childTemplate);
-        resolve(serializedList);
-    });
+    (baseTemplate, childTemplate) => {
+        return Promise.resolve(new Transform(handledTags, insertBeforePipeTags))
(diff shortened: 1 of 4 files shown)


lib/parse-template.js
```diff
@@ -2,9 +2,13 @@
...
-        const serializedList = transform.applyTransforms(baseTemplate, childTemplate);
-        resolve(serializedList);
+    return (baseTemplate, childTemplate) => new Promise((resolve, reject) => {
+        try {
+            const transform = new Transform(handledTags, insertBeforePipeTags);
+            const serializedList = transform.applyTransforms(baseTemplate, childTemplate);
+            resolve(serializedList);
+        } catch (e) {
+            reject(e);
+        }
```
boopathi: Why not 

```js
return () => Promise.resolve(new Transform(a,b)).then(transform => transform.applyTransforms(c,d));
```

here ... `new Transform(a,b)` should not throw ... if it did, it's a library error and we should fix it ... if `applyTransforms` throws then it's a lib-user error, and he might choose to handle it at the end of the chain.

vigneshshanmugam: as you said Transforms should not throw any error and Its handled in the lib itself.. Just to be extra sure to catch errors happening while parsing `templates`, Which shouldn't happen in any case.

lib/serializer.js
```diff
@@ -38,6 +38,19 @@ module.exports = class CustomSerializer extends Serializer {
     }
 
     /**
+     * Gets the serialized content
+     * @returns {String}
+     */
+    getHtmlContent() {
```
mo-gr: is this a one-time-getter? that should be made very clear at least in the jsdoc

vigneshshanmugam: @mo-gr Can you check now..added a better explanation

mo-gr:
:+1:

PR #94, opened 2017-01-03
Allow lazy fragment initializaiton through promise

### LINKED ISSUE #92: Ability to consume promises from init function of fragments
This is helpful if the fragment wants to render lazily or asynchronously.

vigneshshanmugam:
+ This is a backward compatible change

fixes #92

### DIFF
--- a/src/pipe.js
+++ b/src/pipe.js
@@ -49,7 +49,7 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
                 // Exposed fragment initialization Function/Promise
                 var init = i && i.__esModule ? i.default : i;
                 // early return
-                if (!init) {
+                if (typeof init !== 'function') {
                     return;
                 }
                 // check if User Timing API is supported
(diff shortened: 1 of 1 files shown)


src/pipe.js
```diff
@@ -46,21 +46,48 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
...
+                    perf.clearMarks(fragmentId);
+                    perf.clearMarks(fragmentId + '-end');
+                }
+
+                (function executeAndCapture() {
+                    isUTSupported && perf.mark(fragmentId);
+                    // Check if the exposed function is a Promise to allow lazy rendering
+                    // Capture initializaion cost of each fragment on the page using User Timing API if available
+                    if (isPromise(init)) {
+                        init.then(function(module) {
```
grassator: I think that `init` function should rather return a promise

src/pipe.js
```diff
@@ -46,21 +46,46 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
...
-                        perf.mark(fragmentId);
-                        init(node);
+                // early return
+                if (typeof init !== 'function') {
+                    return;
+                }
+                // check if User Timing API is supported
+                var isUTSupported = perf && 'mark' in perf;
+
+                function isPromise(obj) {
```
dazld: pull this out into a library function and import it

vigneshshanmugam: No.. then I would need a built tool and I dont want to go in that path.

dazld: maybe, but it's probably something that should be covered by a test, so pulling it out into a library func is going to help with that.

vigneshshanmugam: Yeah i didnt do that since its not used anywhere other than this one place..

dazld: oki

PR #95, opened 2017-01-04
Provide hooks for measuring performance of fragments on frontend

### LINKED ISSUE #93: Allow a way to measure when the main content becomes interactive
Approach planned

+ mark fragments in the template as **main** - this can be a single / multiple fragments which decides the interactivity of main/meaningful content.
+ `primary` fragments are by default considered as part of main
+ once all the JS assets for fragment marked as main are initialised, we can calculate the interactivity as 
```js
performance.mark('done'); //when all main fragment scripts are executed

performance.measure('interactivity', 'navigationstart', 'done'); //
```

vigneshshanmugam:
+ [x] API Hooks document
+ [x] Performance document
+ [x] example implementation


src/pipe.js
```diff
@@ -74,18 +86,56 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
...
+                    // Handle if there are no main fragments on the page
+                    if (JSON.stringify(mainFragments) === '{}') {
+                        isInteractiveDone = true;
+                        return;
+                    }
+
+                    if (isMainContentInteractive()) {
+                        isInteractiveDone = true;
+                        // This will measure the high resolution time from navigationStart till current time
+                        perf.measure('interactive');
```
dazld: how about `tailor-interactive` instead?

src/pipe.js
```diff
@@ -74,18 +86,56 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
...
+                    }
+                }
+
+                function markInitialized() {
+                    if (typeof mainFragments[index] === 'object') {
```
dazld: would this ever not be an object..?

src/pipe.js
```diff
@@ -74,18 +86,56 @@ var Pipe = (function (doc, perf) { //eslint-disable-line no-unused-vars, strict
...
+                        callback();
+                        // Measure the main content interactivity once all the main fragments are initialized
+                        if (!isInteractiveDone) {
+                            measureInteractivity();
+                        }
+                    };
                     // Check if the response from fragment is a Promise to allow lazy rendering
                     if (isPromise(fragmentRender)) {
-                        fragmentRender.then(callback, callback);
+                        fragmentRender.then(handlerFn, handlerFn);
```
dazld: might be nicer to use the `.catch(handlerFn)` form instead, making it explicit that this fn handles all cases.

src/pipe.js
vigneshshanmugam: yes, when there is are main fragments on the page.

src/pipe.js
vigneshshanmugam: yes much better and less confusing.

doc/Interactivity.md
```diff
@@ -0,0 +1,67 @@
...
+    <fragment main id="product-detail" src="product-detail.zalando.de" />
+    <fragment id="recommendation" src="recos.zalando.de" />
+    <fragment id="footer" src="footer.zalando.de" />
+</body>
+</html>
+```
+
+In the above product page, Header and Product detail fragment are marked as main fragments since these two fragments are enough for the page to provide the critical content to the user. Footer fragment here is below the fold and it is considered not critical.
+
+**Note: The main fragments are maked by the teams that controls the given template/page**
```
dazld: marked*

lib/fragment.js
```diff
@@ -47,7 +48,7 @@ module.exports = class Fragment extends EventEmitter {
     constructor (tag, context, index, requestFragment, pipeInstanceName) {
         super();
         this.attributes = getAttributes(tag, context);
-        ['async', 'primary', 'public'].forEach((key) => {
+        ['async', 'primary', 'public', 'main'].forEach((key) => {
```
dazld: the `primary / main` naming is a little confusing. is there some kind of other name we can give this?

src/pipe.js
dazld: it might swallow errors too.

lib/fragment.js
lmineiro: I agree that 'main' is probably not the best choice

vigneshshanmugam:
Will update the PR with new design discussions and API changes.

lib/fragment.js
dazld: we brainstormed up a better idea - instead of baking in the concept of 'main' timing, to have generic timing groups that are programmatically available to the developer.

vigneshshanmugam:
@dazld @grassator Please have a look and let me know your thoughts.

grassator:
@vigneshshanmugam lgtm

Issue #99, opened 2017-01-09
Migrate code coverage to use Codecov and fix coverage badge


addi90:
The code coverage for tailor is unavailable on [CodeClimate](https://codeclimate.com/github/zalando/tailor/coverage). The coverage badge is also broken because of same.

Use [Codecov](https://codecov.io/) instead to provide a better code coverage reports and graphs

addi90:
I might need contributor permission to authorize codecov. I am checking if I can do without that as well, but I am not sure. I would first implement it in my fork and then raise a PR

PR #100, opened 2017-01-10
#99 Add codecov for integrating code coverage reports

addi90:


addi90:
@mo-gr @vigneshshanmugam Guys the sauce-connect account used for tests does not have a valid auth token. Can you provide me one so that I can update? The tests using sauce-connect are the only ones breaking

package.json
```diff
@@ -12,6 +12,7 @@
     "test": "mocha --harmony tests/**",
     "test-frontend": "mocha --harmony examples/multiple-fragments-with-custom-amd/test.js",
     "coverage": "node --harmony node_modules/.bin/istanbul cover _mocha -- tests/**",
+    "codecov": "cat coverage/*/lcov.info | codecov",
```
addi90: But do we want to club these? If these are isolated we can run the coverage even on local. Thoughts?

vigneshshanmugam:
thanks @addi90 looks good..

PR #103, opened 2017-01-12
Add Link alias note

SimeonC:
I noticed when searching through the code and a co-worker informing me that Tailor has support for AWS' special header format but it wasn't documented anywhere, this commit adds a note to the Readme to highlight this support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ A fragment is an http(s) server that renders only the part of the page and sets 
 
 A JavaScript of the fragment is an AMD module, that exports an `init` function, that will be called with DOM element of the fragment as an argument.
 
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
@SimeonC Thanks for the PR. Yes indeed we kept the format for compatibility reasons and might remove it in the next release.

SimeonC:
@vigneshshanmugam It is useful for us as we are serving our html/js/css files on an AWS cdn and currently have no use case for individual fragment servers, or if you do remove it could you make it an option to change the header name?
(We've adopted tailor for the "microservice" way of splitting the frontend)

vigneshshanmugam:
@SimeonC No issues.. I am happy to keep it incase if you guys are relying on it 👍

PR #109, opened 2017-01-16
Update Tailor using ES6 features(fixes #102)

### LINKED ISSUE #102: Update Tailor to use ES6
Even though NodeJS (along with --harmony flag in v4) provides a lot of ES6 features, the Tailor code doesn't use all of them. Update the code to refactor the Tailor code using ES6. E.g.

1. Add destructuring to files

```javascript
fragmentObj = Object.assign({}, { name: node.name, attributes: node.attribs});
```
to 

```javascript
const { name, attribs: attributes } = node;
fragmentObj = Object.assign({}, { name, attributes});
```

2. Use `() =>` to avoid some boilerplate
```javascript
const stripUrl = (fileUrl) => {
    return path.normalize(fileUrl.replace('file://', ''));
};
```

```javascript
const stripUrl = fileUrl => path.normalize(fileUrl.replace('file://', ''));
```

addi90:
- Use ES6 de-structuring wherever applicable
- Use multiline string templates using \` `
- Use `() => {}` and remove verbose code, wherever applicable
- Use Array methods like `.includes()`, wherever applicable
- Update Tailor to use `node-7.x`


package.json
```diff
@@ -24,7 +24,7 @@
     "test"
   ],
   "engines": {
-    "node": ">4.4.3"
+    "node": ">7.0.0"
```
vigneshshanmugam: > 6.0.0

lib/transform.js
```diff
@@ -47,25 +45,16 @@ module.exports = class Transform {
...
-                this._pushText(node.next, slotMap.get('default'));
+        const slotMap = new Map([['default', []]]);
+        const nodes = treeAdapter.getChildNodes(root);
+        nodes.forEach(node => {
+            if (!treeAdapter.isTextNode(node)) {
+                const { slot = 'default' } = node.attribs;
+                const slotNodes = slotMap.get(slot) || [];
+                const updatedSlotNodes = [...slotNodes, node];
+                slotMap.set(slot, updatedSlotNodes);
+                this._pushText(node.next, updatedSlotNodes);
```
vigneshshanmugam: dont think this works though. `updatedSlotNodes ` doesnt have reference to the slotMap and would not update the text nodes.. I think it should be `slotMap.get(slot)`

vigneshshanmugam:
@addi90 Huge thanks for the PR :)

lib/transform.js
addi90: I think `updatedSlotNodes ` contains the result of line 59-63 which is what is getting used as second parameter of `this._pushText`. So,
```
[...slotNodes, node]  ≈  (slotMap.get(attribs.slot).push(node); || slotMap.set(attribs.slot, [node]);)
```

This is just our understanding of the syntax and seems same in functionality. But, you can confirm if the functionality is broken or not?

lib/fetch-template.js
```diff
@@ -25,30 +26,30 @@ class TemplateError extends Error {
...
 /**
  * Fetches the template from File System
  *
  * @param {string} templatesPath - The path where the templates are stored
  * @param {function=} baseTemplateFn - Function that returns the Base template name for a given page
  */
-module.exports = function fetchTemplate (templatesPath, baseTemplateFn) {
-    return (request, parseTemplate) => {
+module.exports = (templatesPath, baseTemplateFn) => 
+    (request, parseTemplate) => {
```
addi90: I am keeping parenthesis only when returning object. Since , this function returns a promise there is no need of keeping it. Thoughts?

vigneshshanmugam: keeping parentheses for args looks better to me and make you feel its a function.. Its just my thoughts though... If we keep it here,I would also keep it in all arrow functions to be consistent.

lib/transform.js
vigneshshanmugam: I am 99% sure its broken, I will confirm you anyways :)

vigneshshanmugam: I am wrong.. Its working as expected.. sorry for that! :)

addi90: Cool. I was hoping to not have to change this :) @diogodadalt You were right on this one

diogodadalt: @vigneshshanmugam @addi90 thanks for taking the time to review it.

vigneshshanmugam: @diogodadalt No issues man.. I just came to know there should be a test for this as well.. Will add it after this PR has landed.

index.js
```diff
@@ -34,24 +38,27 @@ module.exports = class Tailor extends EventEmitter {
...
-                options.templatesPath ||
+                templatesPath ||
                 path.join(process.cwd(), 'templates')
             ),
             fragmentTag: 'fragment',
             handledTags: [],
             handleTag: () => '',
             requestFragment,
-            pipeDefinition: (pipeInstanceName) => pipeChunk(amdLoaderUrl, pipeInstanceName),
+            pipeDefinition: pipeInstanceName => pipeChunk(amdLoaderUrl, pipeInstanceName),
```
dazld: how about `pipeChunk.bind(null, amdLoaderUrl)` ?

lib/serializer.js
```diff
@@ -64,12 +64,9 @@ module.exports = class CustomSerializer extends Serializer {
      * @returns {Boolean}
      */
     _isSlotNode(node) {
-        if (node.name === 'slot') {
-            return true;
-        }
-        const attribs = node.attribs;
-        return node.name === 'script' &&
-            attribs && attribs.type === 'slot';
+        const { attribs, name} = node;
```
vigneshshanmugam: not sure if eslint rule for this exist but would be cool to add this rule.

lib/request-fragment.js
```diff
@@ -19,8 +19,8 @@ const requiredHeaders = {
  * @param {Object} request - HTTP request stream
  * @returns {Promise} Response from the fragment server
  */
-module.exports = function requestFragment (fragmentUrl, fragmentAttributes, request) {
-    return new Promise((resolve, reject) => {
+module.exports = (fragmentUrl, fragmentAttributes, request) =>
```
dazld: think this should probably stay as a named function expression for stack traces.

vigneshshanmugam: got catch @dazld

lib/fetch-template.js
addi90: Done

lib/transform.js
vigneshshanmugam: added https://github.com/zalando/tailor/commit/f91bcff1f2de64042e448e447040ec0f56831ca5

lib/request-fragment.js
addi90: @dazld   But wouldn't the stack trace use the name from the name of the import. E.g.:
```
const requestFragment = require('./lib/request-fragment');

```

I may be wrong but based on my understanding, with any error, the stack trace would use the name of the import i.e. `requestFragment`.

index.js
addi90: Yeah. We can curry a new function since `pipeInstanceName` is anyway the second param

lib/request-fragment.js
vigneshshanmugam: Maybe you can test it out by throwing error manually?

addi90: Irrespective of the syntax of the function, if its the `default` module function the error thrown is the same for me and defined by the file and line number. E.g. on throwing error in `request-fragment.js`, the below error is thrown with both the syntaxes:

```
if (response.statusCode >= 500) {
    reject(new Error('Internal Server Error'));
} else {
    throw new Error('test the stack trace');
    resolve(response);
}
```

> Uncaught Error: test the stack trace
      at OverriddenClientRequest.fragmentRequest.on (lib/request-fragment.js:43:23)
      at respond (node_modules/nock/lib/request_overrider.js:441:17)
      at node_modules/nock/lib/request_overrider.js:479:11
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickCallback (internal/process/next_tick.js:98:9)

vigneshshanmugam: alright then 👍

addi90: I have fixed the other comment from @dazld and pushed. Check if it makes sense to merge it?

vigneshshanmugam: awesome 👍

lib/serializer.js
addi90: @vigneshshanmugam We can use `http://eslint.org/docs/rules/object-curly-spacing` for this but the problem is the existing code is not consistent for any of the `never` or `always` rules. It throws 58 and 60 errors respectively with `never` and `always` rules. Which one do you think should be used across? I am planning to add this and make it consistent

Issue #104, opened 2017-01-13
Add a fixed set of options to Tailor


vigneshshanmugam:
Tailor currently offers a set of options and every new feature also starts with introducing a new set of options. 

1. Deprecate the unused/unimportant options
2. Cluster the options based on fragment/template level (https://github.com/zalando/tailor/issues/97)
3. Moving completely to ES6 will help in reducing the code bloat when considering the default options as well (https://github.com/zalando/tailor/issues/102)

This is going to be a breaking change and will be introduced as part of the next major release.

PR #110, opened 2017-01-17
configurable options for headers & refactoring [fixes #91, 104]

### LINKED ISSUE #97: Configurable header forwarding for fragment requests
The forwarded headers are currently hard-coded in https://github.com/zalando/tailor/blob/master/lib/filter-headers.js#L16.

This could be configurable via `Tailor(options)` with sane defaults. My particular use case is that tailor runs in a private, trusted network where headers are used for authentication.

I'd be happy to contribute it myself if considered useful.

vigneshshanmugam:
+ Document new options
+ Allow headers to be configurable before forwarding to fragments
+ refactor fragment options. 

fix #97 #104

### DIFF
--- a/lib/filter-headers.js
+++ b/lib/filter-headers.js
@@ -5,15 +5,19 @@ const ACCEPT_HEADERS = ['accept-language', 'referer', 'user-agent'];
  * @callback filterHeaders
  *
  * @param {Object} attributes - Attributes object of the fragment node
- * @param {string} attributes.public - Denotes the public fragemnt. Headers are not forward in this case
+ * @param {string} attributes.public - Denotes the public fragment.
  * @param {Object} request - HTTP Request object
  * @param {Object} request.headers - request header object
  * @returns {Object} New filtered header object
  */
-module.exports = ({ public: publicFragment }, { headers = {} }) =>
-    publicFragment
+module.exports = (attributes, request) => {
+    const { public: publicFragment } = attributes;
+    const { headers = {} } = request;
+    // Headers are not forwarded to public fragment for security reasons
+    return publicFragment
         ? {}
         : ACCEPT_HEADERS.reduce((newHeaders, key) => {
             headers[key] && (newHeaders[key] = headers[key]);
             return newHeaders;
         }, {});
+};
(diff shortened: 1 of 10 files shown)


vigneshshanmugam:
@dazld Please have a look

lib/filter-headers.js
```diff
@@ -2,15 +2,17 @@
...
  * @param {string} attributes.public - Denotes the public fragemnt. Headers are not forward in this case
- * @param {Object} headers - Request header object
+ * @param {Object} request - HTTP Request object
+ * @param {Object} request.headers - request header object
  * @returns {Object} New filtered header object
  */
-module.exports = ({ public: publicFragment }, headers) => 
-    publicFragment 
-        ? {} 
+module.exports = ({ public: publicFragment }, { headers = {} }) =>
```
dazld: Considering this is an example function, is it worth expanding it a bit to make it really explicit what's happening with the request object being passed in? The style is extremely terse!

lib/request-fragment.js
```diff
@@ -14,35 +13,37 @@ const requiredHeaders = {
...
-        const fragmentRequest = protocol.request(options);
-        if (timeout) {
-            fragmentRequest.setTimeout(timeout, fragmentRequest.abort);
-        }
-        fragmentRequest.on('response', (response) => {
-            if (response.statusCode >= 500) {
-                reject(new Error('Internal Server Error'));
-            } else {
-                resolve(response);
+module.exports = (filterHeaders = () => ({})) =>
```
dazld: is it worth putting a default argument here? what's the value? I'd prefer if something was wrong that it fails straight away - ie, throws if it receives undefined for whatever reason.

dazld:
looking good! only a couple of clarity / style comments - feel free to ignore if you like.

lib/filter-headers.js
jmaicher: :+1:

jmaicher: Personal preference:

Since this is the `filterHeaders` function, I would expect the actual headers to filter as a first argument. With the given signature, we always need to extract the headers from the request object (either via argument matching or property access).

Alternative:

```
(headers, attributes, request) => {}
// or
(headers, context) => {} with { attributes, request } = context
```

So if you're not interested in the context, you can simply write:

```
(headers) => {}
```

jmaicher:
LGTM, thanks for the effort :+1:

vigneshshanmugam:
@dazld  Made some modifications as per your suggestions.. 

/cc @jmaicher

dazld:
👍

PR #114, opened 2017-01-19
Upgrade dependencies

vigneshshanmugam:


mo-gr:
:+1:

vigneshshanmugam:
if we have fixed versions, we might end up with upgrading them every now and then when new release spikes in packages..

PR #115, opened 2017-01-20
Handle assertion failures in promises better

mo-gr:
Previously, assertion failures would not have been displayed correctly in the test output.


vigneshshanmugam:
@mo-gr 👍 Cool

w0rm:
@mo-gr @vigneshshanmugam why not return a promise from the tests?

PR #128, opened 2017-02-24
[Perf] Preload primary fragment's assets [fixes #127]

### LINKED ISSUE #127: Proposal - Improving the Interactivity of primary content
Since tailor downloads all the scripts from the fragments asynchronously(Not in order) and all these scripts are loaded as AMD modules by `require.js`, The order of the script execution is not guaranteed since they are `async`. This introduces problems in some cases  

### Problem
+  If the primary fragment script size is too big, Sometimes this does happen in real life and you could see others fragments racing with the primary and getting interactive (scripts being executed) even before `primary`. 

### Ideal workaround
+ `primary` fragments can send a initial code chunk and can do code splitting to lazy load their assets to make it interactive ASAP. 

Even though its the recommended way of improving the performance of the page, It does need lots of iteration and identify the critical `JS` parts that are needed by the page. 

### Preload to the rescue
We could solve this issue by prioritising resources using [Preloading techniques](https://developers.google.com/web/updates/2016/03/link-rel-preload). 

Tailor does wait for the `primary` fragment on any given page since the `primary` fragment decides the Status Code of the Page, We could preload the available assets(JS and CSS) from the fragments before writing the Response Headers of the page. 

Working on the Implementation, Will update the metrics here

vigneshshanmugam:
+ Perf improvements

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -45,6 +45,10 @@ module.exports = function processRequest (options, request, response) {
             return;
         }
         const links = parseLinkHeader(headers.link);
+        // Remove the event listener once we are done capturing the necessary assets
+        // to make sure we are not exceeding EventListener's limit (10 default)
+        this.removeListener('fragment:response', getAssetsToPreload);
+
         // Handle Server rendered fragments without depending on assets
         if (links == null) {
             return;
@@ -55,9 +59,6 @@ module.exports = function processRequest (options, request, response) {
         if (links['fragment-script'] && links['fragment-script'].url) {
             assetsToPreload += `<${links['fragment-script'].url}>; rel="preload"; as="script",`;
         }
-        // Remove the event listener once we are done capturing the necessary assets
-        // to make sure we are not exceeding EventListener's limit (10 default)
-        this.removeListener('fragment:response', getAssetsToPreload);
     };
 
     this.on('fragment:response', getAssetsToPreload);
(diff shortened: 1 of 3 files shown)


lib/request-handler.js
```diff
@@ -35,6 +36,31 @@ module.exports = function processRequest (options, request, response) {
...
+    // Early preloading in browsers to improve the primary fragments
+    const getAssetsToPreload = (fragmentReq, fragment, statusCode, headers) => {
+        // Do not preload assets for non primary fragments
+        if (!fragment.primary) {
+            return;
+        }
+        const links = parseLinkHeader(headers.link);
+        // Handle Server rendered fragments without depending on assets
+        if (links == null) {
+            return;
```
dazld: clean up the listener?

lib/request-handler.js
```diff
@@ -35,6 +36,31 @@ module.exports = function processRequest (options, request, response) {
...
+        }
+        if (links['fragment-script'] && links['fragment-script'].url) {
+            assetsToPreload += `<${links['fragment-script'].url}>; rel="preload"; as="script",`;
+        }
+        // Remove the event listener once we are done capturing the necessary assets
+        // to make sure we are not exceeding EventListener's limit (10 default)
+        this.removeListener('fragment:response', getAssetsToPreload);
+    };
+
+    this.on('fragment:response', getAssetsToPreload);
```
dazld: or make this `once`

dazld: although it gets fired for non primary stuff too.. so this actually wouldn't work.

vigneshshanmugam: Exactly, `once` wont work here

lib/request-handler.js
vigneshshanmugam: good catch

PR #133, opened 2017-03-27
fix uglify options to preserve implicit return

### LINKED ISSUE #131: JS isn't being executed
I've tried to run the `basic-css-and-js` example and i've noticed that the JS isn't taking effect.
Also i get the following errors on the console:
image

vigneshshanmugam:
+ fixes #131 

This breaks our `pipe.min.js` and converts the return statement inside IIFE to a block statement. 

Introduced as part of - https://github.com/mishoo/UglifyJS2/pull/1522

Reminds me that we should use `yarn.lock/shrinkwrap.json` to lock down the versioning of deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint .",
     "fix": "eslint . --fix",
     "benchmark": "node perf/benchmark",
-    "prepublish": "uglifyjs src/pipe.js --mangle --compress=negate_iife=false --comments -o src/pipe.min.js"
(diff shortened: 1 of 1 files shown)


perinikhil:
Thanks a ton @vigneshshanmugam for your speedy support :D
Really appreciate it :)
Also is there anything I can help with on the repo?
I'm looking to make my first open source contribution here 😊

vigneshshanmugam:
@perinikhil Awesome..Glad to hear that.. Anyway you can make the integration tests better? I have already tried to fix here #134

mo-gr:
👍

PR #134, opened 2017-03-28
fix integration test and bump IE version 10

vigneshshanmugam:
Testing async rendering and slots merging to base templates.


mo-gr:
👍

PR #135, opened 2017-03-29
upgrade deps and lock the deps versions

vigneshshanmugam:
+ Lock down the deps using `yarn.lock`

This would help avoid issues like #133 easily in future


examples/basic-css-and-js/index.js
```diff
@@ -10,6 +10,10 @@ const tailor = new Tailor({
 
 // Root Server
 http.createServer((req, res) => {
+    if (req.url === '/favicon.ico') {
+        res.writeHead(200, { 'Content-Type': 'image/x-icon' } );
+        return res.end('');
+    }
```
mo-gr: I like this change, but it has nothing to do with this PR :-)

vigneshshanmugam: True, didn't want to add a separate PR though.. :)

PR #139, opened 2017-03-31
opt out of server push for preloaded resources

vigneshshanmugam:
Disabling the HTTP/2 server push until the day when cache-digests are supported in browsers and all HTTP/2 pushed resources are HTTP cache aware. 

This will also help avoid undesired server behaviour for preloaded resources via `Link` Headers

More details in the SPEC https://w3c.github.io/preload/#server-push-http-2

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -60,11 +60,11 @@ module.exports = function processRequest (options, request, response) {
             return;
         }
         if (links.stylesheet && links.stylesheet.url) {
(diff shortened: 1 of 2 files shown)


mo-gr:
>  until the day when cache-digests are supported in browsers

How do you plan remembering on that day to reenable?

vigneshshanmugam:
> How do you plan remembering on that day to reenable?

I will get an Email :rofl:

Issue #136, opened 2017-03-29
Fragment server with multiple fragment-script / stylesheet resources?


iilei:
Would you appreciate a PR that enables multiple fragment-script / stylesheet resources?

possibly via "name" insted of "rel" property?

vigneshshanmugam:
There is an example here - https://github.com/zalando/tailor/tree/master/examples/multiple-fragments-with-custom-amd

iilei:
I am talking about a different use case. I'd like to send something among the lines of the following:

```
Response Headers:
Link:<http://localhost:8042/js/my-app-libs.js>; rel="fragment-script",<http://localhost:8042/js/my-app-main.js>; rel="fragment-script"
```

The way the headers are parsed in [lib/fragment.js](https://github.com/zalando/tailor/blob/2.1.0/lib/fragment.js#L166-L173), the last definition of a Header Link with `rel="(stylesheet|fragment-script)"` wins: my-app-main.js in this example.

vigneshshanmugam:
@iilei Ahh, Got you. It was a conscious decision to not allow multiple link headers in the response for performance reasons. Fragment Devs could easily bloat the whole page by sending the libs and vendors and whatever chunked scripts as part of the headers. 

The proper way of doing this would be to do code splitting and send the only `main-app.js` that has the necessary modules for the initial state of the application and lazy load all the code splitted modules from the main-app.js

iilei:
Thanks for your response. I get your point. I have made pretty good progress on my fork adding the desired behavior, so if you are interested anyways, I'd be happy to file a PR.

The change would be non breaking and could also be controlled via some default config option (e.g. `{maxScriptLinks : 1, maxStyleLinks : 1, ` or so).

dazld:
Thinking out loud - @vigneshshanmugam and @iilei please do tell me to keep quiet if not helpful - but here's a couple of observations:

- if there is a resource that has already been served, then it shouldn't be included again. I think this provides a number of benefits, in that fragments can be explicit about what they need in terms of resources, without worrying about including the same file multiple times. it boosts resilience and autonomy for fragments as well, in that they don't need to coordinate with other fragments about shared resources, and can just assume that tailor is doing the right thing. I think this is valuable.

[...]

PR #140, opened 2017-04-01
Multiple link rel headers

iilei:
As suggested in #136 - adding an option to allow more than a single Header-Link for css and javascript.

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -3,7 +3,7 @@
 const ContentLengthStream = require('./streams/content-length-stream');
 const EventEmitter = require('events').EventEmitter;
 const PassThrough = require('stream').PassThrough;
-const parseLinkHeader = require('parse-link-header');
+const LinkHeader = require('http-link-header');
 const zlib = require('zlib');
 
 /**
@@ -13,7 +13,7 @@ const zlib = require('zlib');
  * @param {object=} context - Context object for the given fragment
  * @returns {object}
  */
-const getAttributes = (tag, context, index) => {
+const getAttributes = (tag, context) => {
     const attributes = Object.assign({}, tag.attributes);
     const fragmentId = attributes.id;
     if (context && fragmentId && context[fragmentId]) {
@@ -23,7 +23,7 @@ const getAttributes = (tag, context, index) => {
 
     return {
         url: attributes.src,
-        id: fragmentId || index,
+        id: fragmentId,
         fallbackUrl: attributes['fallback-src'],
         timeout: parseInt(attributes.timeout || 3000, 10),
         async: attributes.async,
@@ -51,6 +51,7 @@ module.exports = class Fragment extends EventEmitter {
         index,
         requestFragment,
         pipeInstanceName,
+        maxAssetLinks,
         pipeAttributes = () => {}
     } = {}) {
         super();
@@ -64,8 +65,10 @@ module.exports = class Fragment extends EventEmitter {
             }
             this.attributes[key] = value;
         });
-        this.pipeAttributes = pipeAttributes(Object.assign({}, { id: index }, tag.attributes));
+
         this.index = index;
+        this.maxAssetLinks = maxAssetLinks;
+        this.getPipeAttributes = () => pipeAttributes(Object.assign({}, { id: this.index }, tag.attributes));
         this.requestFragment = requestFragment;
         this.pipeInstanceName = pipeInstanceName;
         this.stream = new PassThrough();
@@ -117,7 +120,10 @@ module.exports = class Fragment extends EventEmitter {
             this.emit('response', response.statusCode, response.headers);
         }
         // Extract the assets from fragment link headers.
-        this.links = parseLinkHeader([response.headers.link, response.headers['x-amz-meta-link']].join(','));
+        const { refs } = LinkHeader.parse([response.headers.link, response.headers['x-amz-meta-link']].join(','));
+
+        this.scriptRefs = refs.filter(ref => ref.rel === 'fragment-script').slice(0, this.maxAssetLinks).map(ref => ref.uri);
+        this.styleRefs = refs.filter(ref => ref.rel === 'stylesheet').slice(0, this.maxAssetLinks).map(ref => ref.uri);
 
         this.insertStart();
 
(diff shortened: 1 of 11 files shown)


lib/request-handler.js
```diff
@@ -22,6 +22,9 @@ const getCrossOriginHeader = (fragmentUrl, host) => {
  */
 module.exports = function processRequest (options, request, response) {
     this.emit('start', request);
+    this.index = 0;
+    this.maxStyleLinks = options.maxStyleLinks;
+    this.maxScriptLinks = options.maxScriptLinks;
```
dazld: unless I'm reading it wrong, these variables are only used in this scope - so they don't really need to be bound to `this`..?

vigneshshanmugam:
@iilei Can you rebase this branch? Will have a look after that. Sorry for that.

iilei:
@vigneshshanmugam did you find the time to look into the rebased branch?

tests/tailor.js
```diff
@@ -800,4 +801,236 @@ describe('Tailor', () => {
...
+                    '<fragment src="https://fragment/2"></fragment>' +
+                    '<fragment src="https://fragment/3"></fragment>'
+                );
+
+            getResponse('http://localhost:8081/test').then((response) => {
+                assert.equal(response.body,
+                    '<html><head></head><body>' +
+                    '<script data-pipe>p.start(0, "http://link-a2", {"id":0})</script>' +
+                    '<script data-pipe>p.start(1, "http://link-a3", {"id":1})</script>' +
+                    '<script data-pipe>p.start(2, "http://link-a4", {"id":2})</script>' +
```
vigneshshanmugam: Can you modify this test with sync and async fragments?

vigneshshanmugam:
@iilei Sorry for the delay in the review, had a vacation. 

I will probably need to check how this broke the performance hooks that we are using to measure the front end script initialisation time of fragments. Will probably need to fix there as well. 

More details here - https://github.com/zalando/tailor/blob/master/docs/Performance.md

tests/tailor.js
```diff
@@ -800,4 +801,236 @@ describe('Tailor', () => {
...
+                    '<html><head></head><body>' +
+                    '<script data-pipe>p.start(0, "http://link3", {"id":0})</script>' +
+                    'hello maxScriptLinks default' +
+                    '<script data-pipe>p.end(0, "http://link3", {"id":0})</script>' +
+                    '</body></html>'
+                );
+            }).then(done, done);
+        });
+
+        it('should handle the last stylesheet Header Link only', (done) => {
```
vigneshshanmugam: IMO, it should only pick the first one and drop everything.

lib/fragment.js
```diff
@@ -64,8 +66,11 @@ module.exports = class Fragment extends EventEmitter {
             }
             this.attributes[key] = value;
         });
-        this.pipeAttributes = pipeAttributes(Object.assign({}, { id: index }, tag.attributes));
+
         this.index = index;
+        this.maxStyleLinks = maxStyleLinks;
```
vigneshshanmugam: Also can we use a single option instead of having two separate option?

tests/tailor.js
iilei: I picked the last one to ensure not to introduce a breaking change. However I am also fine with the first one[s] which is more intuitive

lib/fragment.js
iilei: sure. I'll comment again once it is implemented

vigneshshanmugam: I couldn't come up with a better name. Feel free to propose :)

tests/tailor.js
dazld: first seems to make sense to me too.

tests/tailor.js
iilei: good point. That actually needs work. I'm on it

lib/fragment.js
iilei: Done. I'ts called `maxAssetLinks` for simplicity.

tests/tailor.js
iilei: done

tests/tailor.js
iilei: done. The key was to reverse the calls to Pipe.end for each fragment.

iilei:
@vigneshshanmugam To keep things tidy I squashed the recent changes into a single commit. Regarding the broken performance hooks I am not sure where to look at in order to see what is broken.
Did it get  obsolete by the recent updates perhaps?

vigneshshanmugam:
@iilei Thanks for the rebase, Wouldn't have become obsolete. But we need to modify the way it captures the timings now... It was a single script before, but now its multiple scripts so it will be hard now to figure out the initialisation time. 
Code - https://github.com/zalando/tailor/blob/master/src/pipe.js#L79
Doc - https://github.com/zalando/tailor/blob/master/docs/hooks.md

lib/fragment.js
```diff
@@ -163,26 +169,41 @@ module.exports = class Fragment extends EventEmitter {
...
     }
 
     /**
      * Insert the placeholder for pipe assets at the end of fragment stream
      */
     insertEnd () {
-        if (this.links && this.links['fragment-script']) {
-            this.stream.write(`<script data-pipe>${this.pipeInstanceName}.end(${this.index}, "${this.links['fragment-script'].url}", ${JSON.stringify(this.pipeAttributes)})</script>`);
+        this.index--;
+        if (this.scriptRefs[0]) {
```
vigneshshanmugam: would check for length instead of this.

lib/fragment.js
```diff
@@ -163,26 +169,41 @@ module.exports = class Fragment extends EventEmitter {
...
-        if (this.links && this.links['fragment-script']) {
-            this.stream.write(`<script data-pipe>${this.pipeInstanceName}.start(${this.index}, "${this.links['fragment-script'].url}", ${JSON.stringify(this.pipeAttributes)})</script>`);
-        } else {
+        this.styleRefs.forEach(uri => {
+            this.stream.write(
+                this.attributes.async ? `<script>${this.pipeInstanceName}.loadCSS("${uri}")</script>` : `<link rel="stylesheet" href="${uri}">`
+            );
+        });
+
+        if (!this.scriptRefs[0]) {
```
vigneshshanmugam: same here

examples/multiple-css-and-js/index.js
```diff
@@ -0,0 +1,112 @@
...
+        res.writeHead(200, { 'Content-Type': 'image/x-icon' } );
+        return res.end('');
+    }
+    tailor.requestHandler(req, res);
+}).listen(8080, function () {
+    console.log('Tailor server listening on port 8080');
+});
+
+// Fragment server - Any http server that can serve fragments
+http.createServer((req, res) => {
```
vigneshshanmugam: Also there are lot of examples already and this doesn't need to live under a separate folder, since this is a crtiical featue, I would put this inside `multiple-fragments-with-custom-amd` folder so that we run the integration test https://github.com/zalando/tailor/blob/master/examples/multiple-fragments-with-custom-amd/test.js for the same in sauce labs.

vigneshshanmugam: We can merge this PR once this is done, I have found a way for the performance hooks to work.

vigneshshanmugam:
Sorry it took so long to review properly, we can leave the performance hooks for now. It just needs a small tuning and wont stop tailor's next release.

iilei:
cool. I just started working on a fix for the hooks. Good to know you found a way already 👍

examples/multiple-css-and-js/index.js
iilei: cool. I just started working on a fix for the hooks. Good to know you found a way already 👍

iilei:
@vigneshshanmugam I just did the requested tweaks. Looking good?

vigneshshanmugam:
@iilei Thanks a lot for follow ups and quick fixes around the PR. Sorry, I took a lot of time to review and provide feedbacks.

Lets get this in today, We can do the hooks as the next PR. Thanks a ton! 👍

iilei:
Thanks a bunch! No worries about the delays - very pleasant co-op overall :+1:

PR #141, opened 2017-04-11
fix issue with preloading wrong assets on pages

vigneshshanmugam:
+ The old way of preloading logic is buggy and creates sync issues because it was based on events. 

This is cleaner approach and fixes the issue with preloading wrong assets for previously rendered page. This scenario is hard to test because of the way events work and you would need to loadtest the system to really create this scenario.

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -13,6 +13,24 @@ const getCrossOriginHeader = (fragmentUrl, host) => {
     return '';
 };
 
+// Early preloading of primary fragments assets to improve Performance
+const getAssetsToPreload = ({ link }, { headers = {} }) => {
+    let assetsToPreload = '';
+    const links = parseLinkHeader(link);
+    // Handle Server rendered fragments without depending on assets
+    if (links == null) {
+        return assetsToPreload;
+    }
+    if (links.stylesheet && links.stylesheet.url) {
+        assetsToPreload += `<${links.stylesheet.url}>; rel="preload"; as="style"; nopush,`;
+    }
+    if (links['fragment-script'] && links['fragment-script'].url) {
+        const crossOrigin = getCrossOriginHeader(links['fragment-script'].url, headers.host);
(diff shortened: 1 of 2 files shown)


lib/request-handler.js
```diff
@@ -13,6 +13,24 @@ const getCrossOriginHeader = (fragmentUrl, host) => {
...
+// Early preloading of primary fragments assets to improve Performance
+const getAssetsToPreload = ({ link }, { headers = {} }) => {
+    let assetsToPreload = '';
+    const links = parseLinkHeader(link);
+    // Handle Server rendered fragments without depending on assets
+    if (links == null) {
+        return assetsToPreload;
+    }
+    if (links.stylesheet && links.stylesheet.url) {
+        assetsToPreload += `<${links.stylesheet.url}>; rel="preload"; as="style"; nopush,`;
```
grassator: I think we might end up with a dangling comma here.

vigneshshanmugam: What do you mean? We need the comma between two preloaded assets.

shuhei: Probably a trailing comma when the primary fragment doesn't have a JS?

lib/request-handler.js
```diff
@@ -118,7 +109,10 @@ module.exports = function processRequest (options, request, response) {
                         }
                         this.emit('response', request, statusCode, responseHeaders);
                         // Make resources early discoverable while processing HTML
-                        assetsToPreload && response.setHeader('Link', assetsToPreload);
+                        let preloadAssets = getAssetsToPreload(headers, request);
```
shuhei: This can be `const`.

lib/request-handler.js
shuhei: But it's the existing behaviour.

vigneshshanmugam: Yes and browsers does ignore it.. Don't think we need to do anything here. Thoughts?

lib/request-handler.js
vigneshshanmugam: Oops, thought I had a config for eslint to capture this case.

PR #148, opened 2017-05-12
Forward cookies from first primary fragment.

rikkert:
I'm working on a primary fragment that uses a cookie for its state.
Not sure how you guys deal with this situation.
But I did not see a different way to accomplish this within tailor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ Default implementation [`lib/fetch-template.js`](https://github.com/zalando/tail
 * `fragmentTag` - Name of the fragment tag, defaults to `fragment`
 * `handledTags` - An array of custom tags, check [`tests/handle-tag`](https://github.com/zalando/tailor/blob/master/tests/handle-tag.js) for more info
 * `handleTag(request, tag)` - Receives a tag or closing tag and serializes it to a string or returns a stream
-* `filterHeaders(attributes, request)` - Function that filters the request headers that are passed to fragment request, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
+* `filterRequestHeaders(attributes, request)` - Function that filters the request headers that are passed to fragment request, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
+* `filterResponseHeaders(attributes, response)` - Function that maps the response headers from the primary fragment request to the final response, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
 * `maxAssetLinks` - Number of `Link` Header directives for CSS and JS respected per fragment - defaults to `1`
 * `requestFragment(filterHeaders)(url, attributes, request)` - Function that returns a promise of request to a fragment server, check the default implementation in [`lib/request-fragment`](https://github.com/zalando/tailor/blob/master/lib/request-fragment.js)
 * `amdLoaderUrl` - URL to AMD loader. We use [RequireJS from cdnjs](https://cdnjs.com/libraries/require.js) as deafult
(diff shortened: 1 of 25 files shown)


lib/request-handler.js
```diff
@@ -107,6 +107,9 @@ module.exports = function processRequest (options, request, response) {
                         if (headers.location) {
                             responseHeaders['Location'] = headers.location;
                         }
+                        if (headers['set-cookie']) {
```
vigneshshanmugam: Rather than doing this way, We can have a callback in tailor [options](https://github.com/zalando/tailor#options) and whitelist some of the headers from primary fragment alone. We already have a similar way for filtering headers for the request.

vigneshshanmugam:
Check filter headers implementation here https://github.com/zalando/tailor/blob/master/lib/filter-headers.js

rikkert:
That makes sense, I'll update this PR.

README.md
```diff
@@ -51,7 +51,8 @@ Default implementation [`lib/fetch-template.js`](https://github.com/zalando/tail
 * `fragmentTag` - Name of the fragment tag, defaults to `fragment`
 * `handledTags` - An array of custom tags, check [`tests/handle-tag`](https://github.com/zalando/tailor/blob/master/tests/handle-tag.js) for more info
 * `handleTag(request, tag)` - Receives a tag or closing tag and serializes it to a string or returns a stream
-* `filterHeaders(attributes, request)` - Function that filters the request headers that are passed to fragment request, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
+* `filterRequestHeaders(attributes, request)` - Function that filters the request headers that are passed to fragment request, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
```
vigneshshanmugam: naming ll break the older tailor versions, accept both flilterHeaders and filterRequestHeaders

README.md
```diff
@@ -51,7 +51,8 @@ Default implementation [`lib/fetch-template.js`](https://github.com/zalando/tail
...
+* `filterResponseHeaders(attributes, response)` - Function that maps the response headers from the primary fragment request to the final response, check default implementation in [`lib/filter-headers`](https://github.com/zalando/tailor/blob/master/lib/filter-headers.js)
```
vigneshshanmugam: We dont need to passaround the whole response. Its a huge object. just the header object is enough here.

docs/Events.md
```diff
@@ -15,7 +15,7 @@ Events may be used for logging and monitoring. Check [perf/benchmark.js](https:/
 ## Fragment events
 
 * Request start: `fragment:start(request, fragment.attributes)`
-* Response Start when headers received: `fragment:response(request, fragment.attributes, status, headers)`
+* Response Start when headers received: `fragment:response(request, fragment.attributes, status, headers, response)`
```
vigneshshanmugam: same here

lib/filter-headers.js
```diff
@@ -1,23 +1,41 @@
...
- * @param {Object} request - HTTP Request object
- * @param {Object} request.headers - request header object
- * @returns {Object} New filtered header object
- */
-module.exports = (attributes, request) => {
-    const { public: publicFragment } = attributes;
-    const { headers = {} } = request;
-    // Headers are not forwarded to public fragment for security reasons
+const ACCEPT_REQ_HEADERS = ['accept-language', 'referer', 'user-agent'];
+const ACCEPT_RES_HEADERS = ['location', 'set-cookie'];
```
vigneshshanmugam: passing set-cookie as default will break if people use reverse proxy infront of tailor and set the cookie from there. Keep only location for now. others can override this function if required.

README.md
rikkert: Yes that is true, it does give an inconsistent function signature for: filterRequestHeaders and filterResponseHeaders.

README.md
rikkert: Tried so hard to be backwards compatible, nice catch. :)

lib/filter-headers.js
rikkert: I also realised later that not passing Location for public fragments breaks expected behaviour on a 30x.
Considering this and the fact the function signature will be different I propose to not include a default filter.

vigneshshanmugam:
Thanks 👍 Looks good.

PR #155, opened 2017-06-06
Fix + test hooks

iilei:
* follow-up on #140

### DIFF
--- a/src/pipe.js
+++ b/src/pipe.js
@@ -24,6 +24,53 @@
         }
     }
 
+    function arrayMath(array, mathFn) {
+        return Math[mathFn].apply(null, array);
+    }
+
+    function arrayFlatten(array) {
+        return [].concat.apply([], array);
+    }
+
+    function arrayOf(val, size) {
+        var result = [];
+        for (var i = size; i--; i >= 0) {
+            result[i] = val;
+        }
+        return result;
+    }
+
+    // setChunkInitPhase keeps track of initialization phases
+    function setChunkInitPhase(index, attributes, phase) {
+        var chunkIndex = index - attributes.range[0];
+        initState[attributes.fragmentID][chunkIndex] = phase;
+    }
+
+    function chunkHook(attributes, hook) {
+        var fragmentID = attributes.fragmentID;
+        switch (hook) {
+            case 'onStart':
+                if (!initState[attributes.fragmentID]) {
+                    initState[attributes.fragmentID] = arrayOf(0, attributes.range);
+                    hooks.onStart(attributes);
+                }
+                break;
+            case 'onBeforeInit':
+                var chunkMax = arrayMath(initState[fragmentID], 'max');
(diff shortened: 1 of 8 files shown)


vigneshshanmugam:
Awesome work 👍 , Will do some testing and get back to you. Need to check if this script works as well https://github.com/zalando/tailor/blob/master/examples/fragment-performance/templates/index.html#L7

src/pipe.js
```diff
@@ -24,15 +25,56 @@
         }
     }
 
+    function chunksCount(range) {
+        return (range[1] - range[0] + 1);
```
vigneshshanmugam: you could simply return range.length

src/pipe.js
```diff
@@ -4,6 +4,7 @@
     var scripts = doc.getElementsByTagName('script');
     // State to maintain if all fragments on the page are initialized
     var initState = [];
+    var ranges = [];
```
vigneshshanmugam: Also IMO we can use the existing initState instead of having one more state for checking the initialisation of fragments.

vigneshshanmugam: We can do this by introducing a unique identifier to fragment class. Previously id was unique to a fragment, Now its modified based on the index.

src/pipe.js
iilei: This is only true if maxAssetLinks is 2.
E.g for `var range = [0,9]` it should return `10` and not `2` (`range.length`)

src/pipe.js
```diff
@@ -24,15 +25,56 @@
...
         placeholders[index] = currentScript();
     }
 
     function start(index, script, attributes) {
         starts[index] = currentScript();
         if (script) {
-            initState.push(index);
-            hooks.onStart(attributes);
+            chunkHook(attributes, 'onStart');
+            initState[index] = 'start';
```
vigneshshanmugam: If fragment sends less link headers than the maxAssets link, this will create sparse array and its not good.

src/pipe.js
vigneshshanmugam: Ahh okay. That makes sense!

src/pipe.js
iilei: I can think of two possible ways to deal with the possibility of fragments serving less than maxAssetLinks:

1. sparse arrays
2. objects like `{ fragment0: ['start', 'start' ] , fragment3: ['start', 'start', 'start' ] }` 

I went for the former based on my guess it's executionb would be faster

iilei: Another option would be passing a `fragmentId` along with the `pipeAttributes` ...

iilei: ... actually, that appears to simplify things very much :smiley:

iilei:
I am going to re-open after a bit, needs more work.

iilei:
Just ran into the ["can't reopen PR" issue](https://github.com/isaacs/github/issues/361)

So I am going to file another one. Sorry for the mess 😢

PR #157, opened 2017-06-06
Use the node v8 promisify API to remove unnecessary code

### LINKED ISSUE #153: Use the node v8 promisify API to remove unnecessary code
We can remove some unnecessary explicit `promisify` code by using the v8 `utils.promisify` API (https://nodejs.org/en/blog/release/v8.0.0/#improved-support-for-promises). 

E.g.: This [function](https://github.com/zalando/tailor/blob/master/lib/fetch-template.js#L29) can be changed to a much elegant and declarative implementation.

```javascript
const readFile = (path) => 
    new Promise((resolve, reject) => {
        fs.readFile(path, 'utf-8', (err, data) => {
            if (err) {
                reject(new TemplateError(err));
                return;
            } 
            resolve(data);
        });
    });
```
to

```javascript
const util = require('util');

const readFile = (path) => 
    util.promisify(fs.readFile)(path)
        .catch(err => new TemplateError(err));
```

addityasingh:
- closes #153 
- Use the node 8 `promisify` API to remove unnecessary code
- Polyfill it for older node versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - 6.1.0
   - 7.1.0
+  - 8.0.0
 cache:
   yarn: true
   directories:
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ module.exports = class Tailor extends EventEmitter {
                     memoizedDefinition = `<script src="${amdLoaderUrl}"></script>\n<script>`;
                 }
             }
-            return new Buffer(`${memoizedDefinition}var ${pipeInstanceName}=${PIPE_DEFINITION}</script>\n`);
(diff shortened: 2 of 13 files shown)


addityasingh:
This is auto-generated by `npm 5` to optimise the install time

vigneshshanmugam:
I am aware of that, But we already use `yarn.lock` for determinism.

addityasingh:
Closing this for now, since there are some possible performance issue with the `util.promisify` library

addityasingh:
Seems like wrong PR to me. I will close this and create a new 1

vigneshshanmugam:
The title seems wrong here.  But we need tests for fetch-templates as well :)

PR #159, opened 2017-06-08
Fix hooks

iilei:
* Add tests for Hooks execution order and amount
* follow-up on #140

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -184,11 +184,11 @@ module.exports = class Fragment extends EventEmitter {
         }
 
         const range = [this.index, this.index + this.scriptRefs.length - 1];
+        const fragmentId = this.attributes.id || range[0];
         this.scriptRefs.forEach((uri)=> {
-            const id = this.index;
-            const attributes = Object.assign({}, this.getPipeAttributes(), { id: this.attributes.id || id, range });
+            const attributes = Object.assign({}, this.getPipeAttributes(), { id: fragmentId, range });
             this.stream.write(
-                `<script data-pipe>${this.pipeInstanceName}.start(${id}, "${uri}", ${JSON.stringify(attributes)})</script>`
+                `<script data-pipe>${this.pipeInstanceName}.start(${this.index}, "${uri}", ${JSON.stringify(attributes)})</script>`
             );
             this.index++;
         });
@@ -201,11 +201,11 @@ module.exports = class Fragment extends EventEmitter {
         if (this.scriptRefs.length > 0) {
             const range = [this.index - this.scriptRefs.length, this.index - 1];
             this.index--;
+            const fragmentId = this.attributes.id || range[0];
             this.scriptRefs.reverse().forEach((uri)=> {
-                const id = this.index--;
-                const attributes = Object.assign({}, this.getPipeAttributes(), { id: this.attributes.id || id, range });
+                const attributes = Object.assign({}, this.getPipeAttributes(), { id: fragmentId, range });
                 this.stream.write(
-                    `<script data-pipe>${this.pipeInstanceName}.end(${id}, "${uri}", ${JSON.stringify(attributes)})</script>`
+                    `<script data-pipe>${this.pipeInstanceName}.end(${this.index--}, "${uri}", ${JSON.stringify(attributes)})</script>`
                 );
             });
         } else {
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -171,8 +171,8 @@ module.exports = class Fragment extends EventEmitter {
     insertStart () {
         this.styleRefs.forEach(uri => {
             this.stream.write(
-                this.attributes.async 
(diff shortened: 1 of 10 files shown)


iilei:
I altered the integration test so it checks fragments serving fewer javascript files than permitted by `maxAssetLink` option and also added a test that checks execution of hooks.

lib/fragment.js
```diff
@@ -183,9 +184,12 @@ module.exports = class Fragment extends EventEmitter {
             return;
         }
 
+        const range = [this.index, this.index + this.scriptRefs.length - 1];
+        const fragmentId = this.fragmentId;
+
         this.scriptRefs.forEach((uri)=> {
             const id = this.index;
-            const attributes = Object.assign({}, this.getPipeAttributes(), { id: this.attributes.id || id });
+            const attributes = Object.assign({}, this.getPipeAttributes(), { id: this.attributes.id || id, range, fragmentId });
```
vigneshshanmugam: + We should combine both id and fragmentId 
+ Use index only for script ordering
+ Id/fragmentId for all the other cases - context, perf hooks

vigneshshanmugam: Check here https://github.com/zalando/tailor#fragment-attributes

vigneshshanmugam:
There is an interesting bug here, If a fragment assume `fragment-1` sends three scripts in Link header
```html
Link: <module-1.js>; rel="fragment-script", <module-2.js>; rel="fragment-script", <module-3.js>; rel="fragment-script";
```
1. Module-1 exposes a default export function 
2. Module-2 does not exposes any export function, Instead just executes some arbitrary stuff. 
3. Same as module-1

In this scenario, The hooks except `start` for the fragment-1 will never run and `all-done` will not run as well. 

This was not a problem before because we ignore if the fragment does not expose any default exports and fire all-done at last.

iilei:
Good catch ... I haven't used require.js before and I wonder if there is some callback that would be able to handle the load event of arbitrary stuff.

vigneshshanmugam:
Oops Sadly, this was broken before as well.. 'all-done' was never called if there was a script without a default export. I will try to find some hooks for require.js in the meanwhile.

iilei:
I am also trying to find a way on another branch 'fix_hooks_wip'.
that approach would be less staight forward compared to the theoretical solution incorporating some kind of require callback, though.
let's see what works best in the end.

vigneshshanmugam:
We need to handle three cases

1. if the script is not a AMD module, callback wont even get called
2. if script is AMD and does not have a init function(standalone)
3. if script is AMD and exposes a init fn (handled already - default)

vigneshshanmugam:
I am starting to think, this logic should not be in tailor

1. Tailor should just fire onStart, onBefore, onAfter for all fragment scripts.
2. onDone should be called when all fragment's scripts are done initialising 
3. Whoever wants to hook in to the logic should group them based on fragment Ids or whatever logic. We already handle the timingGroup feature in similar fashion https://github.com/zalando/tailor/blob/master/examples/fragment-performance/templates/index.html#L7

Thoughts? @iilei

vigneshshanmugam:
I will create a new PR with your commits and some of my own.

iilei:
I agree. That would be a good way of separating concerns and reducing the footprint of tailor. therefore :thumbsup:

src/pipe.js
```diff
@@ -63,10 +73,12 @@
         start.parentNode.removeChild(start);
         end.parentNode.removeChild(end);
         script && require([script], function (i) {
-            // Exposed fragment initialization Function/Promise
+            // Exported AMD fragment initialization Function/Promise
             var init = i && i.__esModule ? i.default : i;
             // early return
             if (typeof init !== 'function') {
+                initState.pop();
```
vigneshshanmugam: This will fix the issue when onDone was not called for script without dependencies.

lib/fragment.js
```diff
@@ -197,17 +198,18 @@ module.exports = class Fragment extends EventEmitter {
      * Insert the placeholder for pipe assets at the end of fragment stream
      */
     insertEnd () {
-        this.index--;
         if (this.scriptRefs.length > 0) {
+            const range = [this.index - this.scriptRefs.length, this.index - 1];
+            this.index--;
+            const fragmentId = this.attributes.id || range[0];
```
vigneshshanmugam: By keeping the id consistent, It ll become easy to measure the fragment perf. I will update the fragment-performance script to handle this case.

vigneshshanmugam:
@iilei Its ready, have added some changes to your PR. Let me know your thoughts, We can merge once you review it.

iilei:
@vigneshshanmugam great stuff :+1: 

I just added implicit testing the amount of hook execution by appending semicolons to the log lines.

vigneshshanmugam:
@iilei I did manually remove it :) All good

vigneshshanmugam:
Final rebase done. Merging it

PR #168, opened 2017-08-11
custom hooks should be called for all fragments

vigneshshanmugam:
+ This allows hooks to be called for fragments that does not have a default init function exposed. 

+ Also hoisted the fn creation


mo-gr:
:+1:

vigneshshanmugam:
👍

Issue #170, opened 2017-08-22
Loading a specific template insted of a folder


luizdesign:
I'm trying use a specific template to more than one route like a layout with fragments to header and footer, so I need set a specfic template without relation to request's path.

vigneshshanmugam:
Thanks for creating the issue.

Good catch, Not sure how we missed this part. Will take a look at the PR :)

vigneshshanmugam:
Landed #171

luizdesign:
@vigneshshanmugam Do you have any idea about a new version launch on the npm?

vigneshshanmugam:
Thought of doing it once we land fix for #172 . But if its really urgent I can publish a new version

luizdesign:
Relax man, It's just to bookmark on the our roadmap when we can use this version in production.

vigneshshanmugam:
Cool 👍 

Are you guys using tailor in Production? If so would you be able to link me to the company, website or repo :)

luizdesign:
In fact, we are testing tailor in a concept proof about a new front-end architecture, and making a release plan to launch a mvp. However, we don't define what technology will be used yet.

vigneshshanmugam:
@luizdesign Cool, happy to answer any questions incase you guys need any help. 

Published a new version 3.1.1 with your changes. https://github.com/zalando/tailor/blob/master/CHANGELOG.md#311

luizdesign:
Any others doubts or requests we will open new issues to discuss! =)

Thanks @vigneshshanmugam.

PR #171, opened 2017-08-22
Feature: Allow set specific template

luizdesign:
Proposal to resolve issue #170


luizdesign:
Working to fix broken unit test

package-lock.json
```diff
@@ -0,0 +1,2039 @@
+{
```
vigneshshanmugam: can you remove the package-lock file? we use `yarn.lock`

lib/fetch-template.js
```diff
@@ -26,44 +26,78 @@ class TemplateError extends Error {
...
-        const pathname = url.parse(request.url, true).pathname;
-        const templatePath = path.join(templatesPath, pathname) + '.html';
-
-        return readFile(templatePath)
-            .then((baseTemplate) => {
-                if (typeof baseTemplateFn !== 'function') {
-                    return parseTemplate(baseTemplate);
-                }
+        return getTemplatePath(request, templatesPath)
+            .then((templatePath) => {
```
vigneshshanmugam: can you combine it by this way
```js
getTemplatePath(request, templatesPath)
    .then(readFile)
    .then(baseTemplate)
```
looks cleaner IMO.

vigneshshanmugam:
Thanks a lot for working on this. If you are interested, feel free to use https://github.com/ljharb/util.promisify to promisify the functions. We thought of using that package anyways.

lib/fetch-template.js
```diff
@@ -26,44 +26,78 @@ class TemplateError extends Error {
...
 /**
+ * Returns the template path validating a exactly file or a directory
+ *
+ * @param  {Object} request - Request Object
+ * @param  {String} templatesPath - TemplatesPath config
+ *
+ * @return {Promise} Template Path on success or TemplateError on fail
+ */
+const getTemplatePath = (request, templatesPath) => new Promise((resolve, reject) => {
+    fs.lstat(
```
mroderick: If you're using node 8.x, you can use [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) to make this code simpler

mroderick:
> If you are interested, feel free to use https://github.com/ljharb/util.promisify to promisify the functions. We thought of using that package anyways.

@vigneshshanmugam is [`util.promisify` ](https://nodejs.org/api/util.html#util_util_promisify_original) not available?

vigneshshanmugam:
We still support node 6, so we can use that polyfill for node < 8

lib/fetch-template.js
luizdesign: Done! (:

package-lock.json
luizdesign: Done! (:

lib/fetch-template.js
luizdesign: I followed the code style of 'readFile' function, however we can change the both functions to promisify sintaxe. What do you think? Or do we open a new issue to discuss about it?

vigneshshanmugam: We can also do it in a separate PR.

luizdesign: Ok @vigneshshanmugam, I will open a issue about it.

luizdesign: https://github.com/zalando/tailor/issues/172

lib/fetch-template.js
```diff
@@ -26,35 +26,66 @@ class TemplateError extends Error {
...
  * @param {function=} baseTemplateFn - Function that returns the Base template name for a given page
  */
-module.exports = (templatesPath, baseTemplateFn) => 
+module.exports = (templatesPath, baseTemplateFn) =>
     (request, parseTemplate) => {
-        const pathname = url.parse(request.url, true).pathname;
-        const templatePath = path.join(templatesPath, pathname) + '.html';
-
-        return readFile(templatePath)
+        return getTemplatePath(request, templatesPath)
```
vigneshshanmugam: Two problems with this approach

1. the url parsing happens twice with this approach. In `getTemplatePath` when its a directory and inside `baseTemplateFn`. 
2. You would also need to look at modify this logic https://github.com/zalando/tailor/blob/83446134c35dc5482bc48aa24463e2123d350285/lib/fetch-template.js#L63 Will totally break if the templatesPath is a single file instead of dir.

luizdesign: Interesting, I will check this points as soon as possible!

luizdesign: @vigneshshanmugam I adjust the duplicated url parsing and refactoring the logic to factory the complete template path. Now I working for break the process in case of single file

luizdesign: @vigneshshanmugam Done!

lib/fetch-template.js
```diff
@@ -26,45 +26,100 @@ class TemplateError extends Error {
...
-        const templatePath = path.join(templatesPath, pathname) + '.html';
+        const pathname = getPathName(request);
 
-        return readFile(templatePath)
-            .then((baseTemplate) => {
-                if (typeof baseTemplateFn !== 'function') {
-                    return parseTemplate(baseTemplate);
-                }
+        return getTemplatePath(templatesPath, pathname)
+            .then((templateStat) => {
```
vigneshshanmugam: just a small change to keep the indentation depth minimal  
```js
.then({path} => readFile(path))
.then(baseTemplate => ....)
```

lib/fetch-template.js
vigneshshanmugam: Looks good to me 👍

lib/fetch-template.js
luizdesign: @vigneshshanmugam I understand your point of view, however i use the templateStat object to break the template parse logic on resolve the readFile promise. In your logic this not working apparently:
```js
.then(({path}) => readFile(path))
.then(baseTemplate => {
    // templateStat is not defined
})
```

We can pass the complete object to readFile, add the baseTemplate to object and resolve the promise with the object. What do you think about it?

vigneshshanmugam: Ahh thanks for pointing out. No this is fine then, We can do the cleanup as part of using `util.promisify` library. 
lets get this merged :)

vigneshshanmugam:
👍

addityasingh:
👎

addityasingh:
👍

PR #173, opened 2017-08-25
Extract tag handling from request handling

DeTeam:
**Motivation**

Right now in tailor `request-handler.js` contains quite a lot of logic. It's responsible for several things:

* Kicking of context and template fetching
* Assembling of the tag-handling logic
* Deciding on how to handle various events in the app (both in the result stream and in the primary fragment)
* Takes care of async stream management

With so many responsibilities it's getting harder to allow more flexible composition to users of tailor. One particular use-case: we want to introduce a new tag, which itself would result into some template which has to be processed in the similar way. `handleTag` function already allows returning streams, but it's quite an overhead to duplicate existing handling algorithm.

**Solution**: extract tag/template handling logic and make it composable. Right now it's extracted under `buildBlock` which can be used inside your own implementation of `handleTag`, so we allow recursion.

Things which were standing in the way during implementation:

* `index` management and `asyncStream` 
* `shouldWriteHead` was kind of everywhere
* `request` object is all over the place + kinda need to separate tailor context (the one overriding fragment info) and more broad term context (some state which is being modified during template processing, for instance, index) 

TODOs:

- [x] Separate tag-handling logic (working name - "block", though I don't really like it, mb we can come up with smth better)
- [x] Add support for per-fragments contexts (this is not directly related, so we might consider make a separate PR with it)
- [x] Make sure parsing template within `buidBlock` works properly
- [x] Extract and revert example changes
- [x] Replace package lock with yarn lock
- [x] Complete the unit tests
- [x] Check performance impact
- [x] Adjust the documentation


DeTeam:
👎 

Still in progress!

examples/custom-tags/templates/index.html
```diff
@@ -0,0 +1,4 @@
+<script type="text/javascript"></script>
```
DeTeam: Without it, the pipe script would be inserted after the switcher and won't be available to it's content. @vigneshshanmugam mb you know the nice way to do it?

vigneshshanmugam: Yeah pipe is inserted before `fragment` tag or one of the tags in handleTags list. Just pass them in the example :)

DeTeam: You mean `pipeTags`?

package-lock.json
```diff
@@ -0,0 +1,2754 @@
+{
+  "name": "node-tailor",
```
vigneshshanmugam: we use `yarn.lock`, can you update yarn.lock instead?

DeTeam: Yes, will do!

index.js
```diff
@@ -64,7 +64,7 @@ module.exports = class Tailor extends EventEmitter {
 
         requestOptions.parseTemplate = parseTemplate(
             [requestOptions.fragmentTag].concat(requestOptions.handledTags),
-            ['script', requestOptions.fragmentTag]
+            ['script', requestOptions.fragmentTag].concat(requestOptions.handledTags)
```
DeTeam: @vigneshshanmugam I guess you were meant this. Those custom `handledTags` where not passed to pipe tags, so for my `switcher` example it was not added. Not sure it's the best way, though.

package-lock.json
DeTeam: There were no updates, I just happened to use latset npm & node %)

index.js
vigneshshanmugam: Yeah you are right. Just feels like this gives too many options of having where pipe should be inserted. Actually, pipe is never required if its not a fragment, "script" was added because anyone can add fragment in the head like `<script src="blah" type="fragment"/>`.

I would move the `fragment` in your example before `switcher` to make things easier.

DeTeam: Or have a script tag :)
It was an interesting side-effect I didn't think about.
People who implement their own tags should kinda be aware then :)

package-lock.json
DeTeam: Done

examples/custom-tags/index.js
```diff
@@ -0,0 +1,77 @@
...
+        return res.end('');
+    }
+    tailor.requestHandler(req, res);
+}).listen(8080, function () {
+    console.log('Tailor server listening on port 8080');
+});
+
+// Fragment server - Any http server that can serve fragments
+http.createServer((req, res) => {
+    const urlObj = require('url').parse(req.url, true);
```
vigneshshanmugam: can you move the require to the top.

index.js
vigneshshanmugam: Yeah true, lets remove these options and insert a empty div tag like before.

lib/request-handler.js
```diff
@@ -63,94 +63,100 @@ module.exports = function processRequest (options, request, response) {
...
+                response.end();
             }
+        } else {
+            contentLengthStream.end();
+        }
+    };
+
+    Promise.all([
+        templatePromise,
+        contextPromise
```
vigneshshanmugam: an error in context should not control the page and return 500.

lib/block.js
```diff
@@ -0,0 +1,147 @@
...
+// context -> quilt context -> fragment overrides
+// request -> original request info
+module.exports.buildBlock = function buildBlock(request, originalContext) {
+    // NOTE: should there be a separation between context as something global
+    // vs. context as quilt/tailor-specific term?
+    let context = originalContext;
+
+    const resultStream = new StringifierStream((tag) => {
+
+        const {
```
vigneshshanmugam: this destructuring logic happens on every invocation of tag, should be moved out of of this function

lib/block.js
```diff
@@ -0,0 +1,147 @@
...
+
+                const pipeFragmentResult = result => {
+                    if (result instanceof stream) {
+                        result.pipe(tempStream);
+                    } else {
+                        tempStream.end(result);
+                    }
+                };
+
+                fetchDynamicContext(tag)
```
vigneshshanmugam: IMO this creates unnecessary complexity in tailor itself and can be handled outside in requestFragment logic. 

Tailor itself is extendable, we may not need this feature here.

DeTeam: Good thoughts. It doesn't really belong to this PR.

lib/request-handler.js
DeTeam: There's a catch above.. Not super visible, though %)

lib/request-handler.js
```diff
@@ -63,94 +63,100 @@ module.exports = function processRequest (options, request, response) {
...
+
+                response.writeHead(statusCode, responseHeaders);
+                blockStream
+                    .pipe(contentLengthStream)
+                    .pipe(response);
+            });
+
+            fragment.on('fallback', (err) => {
+                this.emit('error', request, err);
+                // TODO: add some responseHeaders?
```
vigneshshanmugam: why not `responseHeaders` like before? Do you have anything specific in mind? 

Its also configurable through `filterResponseHeaders` function

lib/request-handler.js
```diff
@@ -63,94 +63,100 @@ module.exports = function processRequest (options, request, response) {
...
+                response.writeHead(500, {});
+                blockStream
+                    .pipe(contentLengthStream)
+                    .pipe(response);
+            });
+
+            fragment.on('error', (err) => {
+                this.emit('error', request, err);
+                // TODO: add some responseHeaders?
```
vigneshshanmugam: Same here.

lib/request-handler.js
```diff
@@ -63,94 +63,100 @@ module.exports = function processRequest (options, request, response) {
...
+        templatePromise,
+        contextPromise
+    ]).then(([ parsedTemplate, context ]) => {
+        const blockStream = buildBlock(request, Object.assign({
+            index: 0,
+            asyncStream
+        }, options, context));
+
+
+        FRAGMENT_EVENTS.forEach((eventName) => {
```
vigneshshanmugam: assuming these events are for fragments. Could you explain me why it is required here?

tests/serializer.js
```diff
@@ -9,6 +9,8 @@ describe('Serializer', () => {
     const serializerOptions = {
         treeAdapter: adapter,
         slotMap: new Map(),
+        addPlaceholders: true,
```
vigneshshanmugam: don't think its used anywhere.

lib/block.js
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            return fragment.fetch(request, false);
+        };
+
+        if (placeholder === 'pipe') {
+            return pipeDefinition(pipeInstanceName);
+        }
+
+        if (asyncStream && placeholder === 'async') {
```
vigneshshanmugam: don't think we need this `asyncstream` check. Its gonna be present all the time.

As an enhancement we can do this
+ create `asyncstream` and placeholders only when async fragments are present in the template.

lib/request-handler.js
DeTeam: It's part of the tailor API (fragment:start, fragment:end, etc. events.) for fragments. It was there before (almost exact same code), difference is that now instead of subscribing to the fragment directly we subscribe to the "block" events (it propagates them), cause fragments are not directly exposed to the request handler.

Does this make sense?

lib/block.js
```diff
@@ -0,0 +1,102 @@
...
+    // vs. context as quilt/tailor-specific term?
+    let context = originalContext;
+
+    const {
+        maxAssetLinks,
+        asyncStream,
+        handleTag,
+        requestFragment,
+
+        // Should the two below be one???
```
vigneshshanmugam: can you explain what you mean here? having a single option instead of 3 separate options?

lib/block.js
```diff
@@ -0,0 +1,102 @@
...
+            });
+
+            FRAGMENT_EVENTS.forEach((eventName) => {
+                fragment.on(eventName, (...args) => {
+                    const prefixedName = 'fragment:' + eventName;
+                    resultStream.emit(prefixedName, request, fragment.attributes, ...args);
+                });
+            });
+
+            const isAsync = fragment.attributes.async;
```
vigneshshanmugam: shall we keep the code similar to how it was in request-handler? 

https://github.com/zalando/tailor/blob/7d0fff28be95c811af064c7d3290e36badc6a419/lib/request-handler.js#L105

lib/block.js
```diff
@@ -0,0 +1,102 @@
...
+            resultStream.emit('fragment:found');
+
+            return fetchFragment(fragmentContext);
+        }
+
+        let result;
+        try {
+            result = handleTag(request, context, tag);
+        } catch (error) {
+            console.error(error);
```
vigneshshanmugam: should we handle the error here? isn't already wrapped in request-handler in catch statement

lib/block.js
```diff
@@ -0,0 +1,102 @@
...
+
+        if (placeholder === 'async') {
+            // end of body tag
+            return asyncStream;
+        }
+
+        if (name === 'fragment') {
+            // Freeze current context & update the indexes
+            const fragmentContext = context;
+            context = Object.assign({}, context, {
```
vigneshshanmugam: Since the only changing part here is index, can we have a just update index instead of doing Object overrides.

lib/block.js
```diff
@@ -0,0 +1,102 @@
...
+            result = handleTag(request, context, tag);
+        } catch (error) {
+            console.error(error);
+        }
+
+        if (result instanceof stream) {
+            result.on('fragment:found', () => {
+                resultStream.emit('fragment:found');
+
+                context = Object.assign({}, context, {
```
vigneshshanmugam: same here.

lib/request-handler.js
vigneshshanmugam: cool. missed it :)

lib/request-handler.js
```diff
@@ -63,94 +63,97 @@ module.exports = function processRequest (options, request, response) {
...
+            contentLengthStream.end();
+        }
+    };
+
+    Promise.all([
+        templatePromise,
+        contextPromise
+    ]).then(([ parsedTemplate, context ]) => {
+        const blockStream = buildBlock(request, Object.assign({
+            index: 0,
```
vigneshshanmugam: don't allow useland code to update index at any point.

lib/request-handler.js
vigneshshanmugam: Yeah cool, looks a bit bad that we need to handle it twice.

perf/benchmark.js
```diff
@@ -22,8 +23,24 @@ const parseTemplate = require('../lib/parse-template')(['fragment'], []);
 const requests = new WeakMap();
 const fragments = new WeakMap();
 const Tailor = require('../');
+const buildBlock = require('../lib/block');
+
+const handleTag = (request, context, tag) => {
+    if (tag && tag.name === 'switcher') {
+        const stream = buildBlock(request, context);
+        process.nextTick(() => {
```
vigneshshanmugam: is it required? `buildBlock` a sync function

lib/block.js
DeTeam: Yeah, it's a note to myself: we have many pipe-related options, I was wondering if some of the options can be joined (the function part with the attribute part, cause we just pass one to another :D )

lib/block.js
DeTeam: Agreed! 

Wow, how can you comment with code snippets like this? oO

lib/block.js
DeTeam: I think we can remove the try-catch. If this throw, `error` event should be emitted on the stringifier stream and get caught in the request-handler.

lib/request-handler.js
DeTeam: Wow, you're right. Thanks for catching!

perf/benchmark.js
DeTeam: I was thinking about kinda real-time example when data is being pushed into this "block" sometime later on (but without big impact on performance).

DeTeam:
Also, let's rename `buildBlock` to something like:

- `processTemplate`
- `handleTemplate`
- other options?

\* template should be parsed into a set of Buffers + Objects

lib/block.js
DeTeam: Actually, not sure here, I pass the context which was there at the time we found the tag further down the line. If I mutate it, fragment might be having some incorrect data inside and then I'd mess up indexes :(

lib/block.js
vigneshshanmugam: hehe.. there is a small option at the left end where you can copy the preview link once you click on the line.

perf/benchmark.js
vigneshshanmugam: Hmmm true, then lets introduce artificial delay and check if back pressure is handled properly.. `setTimeout` would suffice.

vigneshshanmugam:
parseTemplate sounds good to me.

DeTeam:
@vigneshshanmugam we already have `parse` coming from parse5, it's a bit different, I'd stick to `processTemplate`

DeTeam:
Rename is done, `processTemplate` is there!

DeTeam:
Now I also need to check if it make sense to split context and options again

DeTeam:
Rebased. No conflicts. Now going to apply options-context changes.

tests/handle-tag.js
```diff
@@ -42,7 +42,7 @@ describe('Handle tag', () => {
         mockHandleTag.returns('');
         http.get('http://localhost:8080/template', (response) => {
             const request = mockHandleTag.args[0][0];
-            const tag = mockHandleTag.args[0][1];
+            const tag = mockHandleTag.args[0][3];
```
vigneshshanmugam: can we keep the tag as arg1 .. otherwise it wont be backwards compatible.

DeTeam: aaaaah
good point
yes

tests/process-template.js
```diff
@@ -0,0 +1,127 @@
...
+                done();
+            });
+            resultStream.resume();
+            options.asyncStream.end();
+            resultStream.write({ placeholder: 'pipe' });
+            resultStream.write({ name: 'fragment', attributes: { id: 'f2', primary: true, src: 'http://fragment/f2' } });
+            resultStream.write({ placeholder: 'async' });
+            resultStream.end();
+        });
+        it('insert pipe definition in the beginning');
```
vigneshshanmugam: once we finish the tests for process-template we can go ahead and merge it.. rest looks good to me..

DeTeam: Also seem like we're not changing the API much here. Mb minor version update would be suitable.

tests/process-template.js
```diff
@@ -118,10 +118,55 @@ describe('processTemplate', () => {
...
+        });
+        it('handles indexes correctly', (done) => {
+            let data = '';
+
+            function assertIndex(index, fragmentText) {
+                const r = new RegExp(`TailorPipe[.]start[(]${index}[,)].+${fragmentText}.+TailorPipe[.]end[(]${index}[,)]`, 'g');
+                const doesMatch = r.test(data);
+                let message = `No match for the fragmend(index: ${index}, text: ${fragmentText})`;
+                if (!doesMatch) {
+                    console.log(r, data);
```
vigneshshanmugam: should we throw error here?

vigneshshanmugam:
👍

DeTeam:
:+1:

PR #182, opened 2017-09-11
remove uncessary check for parse result

vigneshshanmugam:
+ the serialised parsed template is always an array.

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -186,14 +186,11 @@ module.exports = function processRequest (options, request, response) {
 
         resultStream.once('error', handleError);
 
-        if (Array.isArray(parsedTemplate)) {
-            parsedTemplate.forEach((item) => {
-                resultStream.write(item);
-            });
-            resultStream.end();
-        } else {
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
👍

DeTeam:
:+1:

lib/request-handler.js
```diff
@@ -186,14 +186,11 @@ module.exports = function processRequest (options, request, response) {
...
 
-        if (Array.isArray(parsedTemplate)) {
-            parsedTemplate.forEach((item) => {
-                resultStream.write(item);
-            });
-            resultStream.end();
-        } else {
-            resultStream.end(parsedTemplate);
-        }
+        parsedTemplate.forEach((item) => {
```
addityasingh: @vigneshshanmugam Would it make sense to add a test or a task to add test for this module?

vigneshshanmugam: Its already there in tests/parse-template.. since we never reached this part, codecov was decreased. I found this only through codecov report.

PR #185, opened 2017-09-21
End asyncStream later in the process (before piping)

DeTeam:
Right now asyncStream ends as soon as we stop pushing things to the stringifier stream, which means we can't push things to asyncStream from our customer tags in handleTag.

To support async fragments there we need to end asyncStream later. It seem like proper place would be to end it right before piping it into the resulting stream. I could either do it in the stringifier stream, but though that it'd be better to do via events (`plugged` event).

### DIFF
--- a/lib/streams/stringifier-stream.js
+++ b/lib/streams/stringifier-stream.js
@@ -22,6 +22,8 @@ module.exports = class StringifierStream extends stream.Transform {
         }
         let st = this.queue.shift();
         if (st instanceof stream) {
+            st.emit('plugged');
+
             st.setMaxListeners(st.getMaxListeners() + 1);
             this.isBusy = true;
 
--- a/lib/streams/stringifier-stream.js
+++ b/lib/streams/stringifier-stream.js
@@ -1,5 +1,6 @@
 'use strict';
 const stream = require('stream');
+const AsyncStream = require('./async-stream');
 
 module.exports = class StringifierStream extends stream.Transform {
     constructor(fn) {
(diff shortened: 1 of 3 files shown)


lib/streams/stringifier-stream.js
```diff
@@ -22,6 +22,8 @@ module.exports = class StringifierStream extends stream.Transform {
         }
         let st = this.queue.shift();
         if (st instanceof stream) {
+            st.emit('plugged');
```
vigneshshanmugam: would only emit this if its a instance of async stream though. Totally useless for other fragment streams.

vigneshshanmugam:
Also could you add a test with the issue so that its easier to track why we moved re-emit it again.

DeTeam:
Sure thing, on it!

lib/streams/stringifier-stream.js
DeTeam: Currently yes, but I didn't want to mix up knowledge of async stream into the stringifier stream

vigneshshanmugam: In that case we would need alternative way of handling for this particular instance.

tests/tailor.js
```diff
@@ -56,6 +57,24 @@ describe('Tailor', () => {
...
             },
+            handledTags: ['delayed-fragment'],
+            handleTag: (request, tag, options, context) => {
+                if (tag.name === 'delayed-fragment') {
+                    const st = processTemplate(request, options, context);
+                    setTimeout(() => {
+                        st.end({
+                            name: 'fragment',
+                            attributes: {
+                                async: true,
```
DeTeam: Before, since the fragment's async + also located in handleTag, there would be an error, cause one can't write to async channel after it's ended (which happens on 'finish' events which is all fine, cause we stop writing to the original resultStream right after pushing all parsed chunks).

DeTeam:
:+1:

vigneshshanmugam:
👍

PR #191, opened 2017-10-19
Tailor fails to parse template with comments inside

DeTeam:
For some reason tailor fails to parse templates with HTML comments inside.

Judging by the error message and some snippets in the code, it seem like handling of comments is different from how other tags are processed.

```
     TypeError: Cannot destructure property `slot` of 'undefined' or 'null'.
      at nodes.forEach.node (lib/transform.js:56:51)
      at Array.forEach (<anonymous>)
      at Transform._groupSlots (lib/transform.js:54:15)
      at Transform.applyTransforms (lib/transform.js:32:24)
      at Promise.resolve.then.transform (lib/parse-template.js:19:19)
      at <anonymous>
```

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -52,7 +52,10 @@ module.exports = class Transform {
         const slotMap = new Map([['default', []]]);
         const nodes = treeAdapter.getChildNodes(root);
         nodes.forEach(node => {
-            if (!treeAdapter.isTextNode(node)) {
+            if (
+                !treeAdapter.isTextNode(node) &&
+                !treeAdapter.isCommentNode(node)
+            ) {
                 const { slot = 'default' } = node.attribs;
                 const slotNodes = slotMap.get(slot) || [];
                 const updatedSlotNodes = [...slotNodes, node];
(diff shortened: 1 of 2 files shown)


ruiaraujo:
:+1:

DeTeam:
:+1:

lib/transform.js
```diff
@@ -52,7 +52,10 @@ module.exports = class Transform {
         const slotMap = new Map([['default', []]]);
         const nodes = treeAdapter.getChildNodes(root);
         nodes.forEach(node => {
-            if (!treeAdapter.isTextNode(node)) {
+            if (
+                !treeAdapter.isTextNode(node) &&
+                !treeAdapter.isCommentNode(node)
```
vigneshshanmugam: should not be handled in this way.. 

This fix mean would completely skip comment nodes from child templates which is not the expected output.

ruiaraujo: ok, so why do we handle text nodes like?

Tomorrow I will take a look on passing it through.

DeTeam: > ok, so why do we handle text nodes like?

yup, that was a bit puzzling, thus was thinking it's ok to have comments handled the same way

vigneshshanmugam: Totally forgot why i did it that way. Will update once i remember. 

But, with the fix #195, we can even include text nodes(space) as part of child templates.

Issue #196, opened 2017-11-09
How to properly handle dependencies


tsnolan23:
Sorry if this isn't the place to ask this, I know it isn't actually an issue, but I couldn't find anywhere else to ask it or find information on it.   

**What is the proper/recommended way of handling dependencies within a template between different fragments?**

I've watched most of the Mosaic presentations for Zalando employees, but I'm struggling to see how allowing each team to choose their own stack doesn't lead to dependency overload and poor performance with this architecture.  For example, say I have 4 fragments composing a template. Two of them use the same version of React, one uses Vue, and another uses an older version of React. 

Based on the described architecture, the above scenario isn't exactly far-fetched - teams choose their stack and may be using different versions of the same library, so how do you prevent the browser from needing to load 2 different versions of React and Vue from the client-side?  

Is the idea that you should always "try" to be consistent with the stack, unless otherwise necessary? Mixing stacks at any level require all those dependencies to be run in the browser which means based on the template, you may be loading several MVC frameworks just to support each individual fragment.

I saw a thread (#160) that mentioned that you use RequireJS internally for managing the dependencies, so do you just require everyone that uses React to use the same version?

I'm also struggling to see how that method aligns to the architecture because in one of the lectures it was mentioned that the team that owns the template, doesn't need to own the fragment and vice-versa, so how would a team know which dependencies to include if it's not included in the bundle?

I've just started playing around with this architecture, so sorry if this is listed somewhere, but I'm excited to try and build a prototype, but running into some of the above and would welcome any information you could provide!

If there is a more proper location to discuss this outside of the GitHub issues, I'd more than happy to move it there.

mo-gr:
This is absolutely the right place to ask.

mo-gr:
When we created mosaic, we tried very hard to keep it agnostic of specific frontend frameworks. But the reality of using mosaic is, that a mosaic installation will very likely dictate a dominant framework. In our case, we highly recommend to the teams to use react and we have a fragment in every base template in tailor, that provides a small environment, that the fragments are allowed to rely on. In our case, those are some polyfills, a communication bus and a clearly defined version of React.

In general, nothing stops a fragment from bundling their own version of angular or similar. But in our opinion, there is very little value in having multiple big frameworks on the same page at the same time. So we strongly discourage it.

So your understanding is correct, every fragment just bundles whatever it needs in its fragment script and purely by convention, we discourage teams from using big/heavy frameworks or libraries, that fail if they are included on the same page twice.

[...]

PR #200, opened 2017-11-27
Include demo SPA repos in README

tsnolan23:
The examples provided by default through Tailor are great to get started, but I think linking directly to the SPA examples in the README would be helpful to those looking to dive a bit deeper into more complex and practical usage of Tailor (similar to the techniques used by Zalando in production).

The second of the two example SPAs was built in response to a dependency handling issue (#196)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,13 @@ Go to [http://localhost:8080/index](http://localhost:8080/index) after running t
 
 **Note: Please run the examples with node versions > 6.0.0**
 
+***
+
+Single-page application examples are also available:
+
+* https://github.com/vigneshshanmugam/tailor-spa
+* https://github.com/tsnolan23/tailor-react-spa
+
 ## Benchmark
 
 To start running benchmark execute `npm run benchmark` and wait for couple of seconds to see the results.


vigneshshanmugam:
Thanks @tsnolan23 for the PR :)

vigneshshanmugam:
👍

DeTeam:
:+1:

stevoPerisic:
@tsnolan23, @vigneshshanmugam how would one put together a SPA with the use of fragments? This is not very clear as SPAs navigate via routes and the HTML is rendered on the front-end not on the server. I hope you see my dilemma, if I change routes within a SPA and want to navigate to another fragment how would I execute that?

tsnolan23:
@vigneshshanmugam I know this has been closed, but @stevoPerisic makes a fair point. I'm also curious on that topic.

In certain instances, you're essentially "forfeiting" the true SPA behavior in terms of routing, no? If the fragment uses say, `react-router` to navigate within itself, it works fine, but in a case where the page is completely changing, it's an entirely new HTTP request through skipper/tailor, to get the new template correct? 

It's not like code splitting where it would just make an asynchronous call for the next fragment?

PR #203, opened 2017-12-17
Prioritize the module loader script with link headers

### LINKED ISSUE #202: preload the module loader script to avoid getting deprioritized
+ If fragment teams start using preload headers, the module loader script is pushed back even though its a synchronous script and executed before everything. 

+ Preloading the module loader (requireJS) will solve this since we can stream it to the browser when we open then `head` tag.

vigneshshanmugam:
+ Fixes #202 

+ While working on this feature, figured out a Chrome bug which causes the module loader script in preload cache to be ignored. 

screen shot 2017-12-14 at 14 38 49

Link to the bug - bugs.chromium.org/p/chromium/issues/detail?id=794958

+ With this fix, loader script should be the first to be requested, Check before and after

**Before**
screen shot 2017-12-17 at 18 52 55

**After**
screen shot 2017-12-17 at 18 51 46

Works only in prime cache because of the chrome bug.

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -1,7 +1,16 @@
 'use strict';
+
 const AsyncStream = require('./streams/async-stream');
-const LinkHeader = require('http-link-header');
 const ContentLengthStream = require('./streams/content-length-stream');
+const { TEMPLATE_NOT_FOUND } = require('./fetch-template');
+const processTemplate = require('./process-template');
+const {
+    getLoaderScript,
+    getFragmentAssetsToPreload,
+    nextIndexGenerator
+} = require('./utils');
+
+// Events emitted by fragments on the template
 const FRAGMENT_EVENTS = [
     'start',
     'response',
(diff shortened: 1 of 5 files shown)


tests/tailor.js
```diff
@@ -361,30 +382,11 @@ describe('Tailor', () => {
...
                     assert.equal(
                         response.headers.link,
-                        '<http://localhost:8080>; rel="preload"; as="style"; nopush,<http://localhost:8080>; rel="preload"; as="script"; nopush; '
+                        '<http://loader>; rel="preload"; as="script"; nopush; crossorigin,<http://localhost:8080>; rel="preload"; as="style"; nopush;,<http://localhost:8080>; rel="preload"; as="script"; nopush;'
                     );
                 })
                 .then(done, done);
         });
-
-        it('"shouldn\'t send preloading headers if primary fragment doesn\'t exist"', done => {
```
vigneshshanmugam: test covered in https://github.com/zalando/tailor/pull/203/files#diff-67c49567f86ec6fe981f9b76efc15c75R319

tests/tailor.js
```diff
@@ -249,7 +249,6 @@ describe('Tailor', () => {
                 .then(response => {
                     assert.equal(response.statusCode, 200);
                     assert.deepEqual(response.headers['set-cookie'], [cookie]);
-                    assert.equal(response.headers.link, 'http://link');
```
vigneshshanmugam: Disallow primary fragments from overriding link headers since primary fragments have no context of module loading and piping mechanism.

lib/utils.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return '';
+};
+
+const getPreloadAttributes = ({
+    assetUrl,
+    host,
+    asAttribute,
+    corsCheck = false,
+    noPush = true // Disable HTTP/2 Push behaviour for now
```
vigneshshanmugam: Will be enabled in future once cache digests spec is implemented in browsers.

tests/tailor.js
vigneshshanmugam: We are also including their assets after the loader asset.

lib/request-handler.js
```diff
@@ -142,12 +104,16 @@ module.exports = function processRequest(options, request, response) {
...
-                responseHeaders.link = preloadAssets;
-            }
+            const assetsToPreload = getFragmentAssetsToPreload(
+                headers,
+                request.headers
+            );
+
+            // Loader script must be preloaded before every fragment asset
+            const loaderScript = getLoaderScript(amdLoaderUrl, request.headers);
+            loaderScript !== '' && assetsToPreload.unshift(loaderScript);
```
mo-gr: any reason not to use explicit `if () {}` here?

lib/request-handler.js
```diff
@@ -202,7 +170,18 @@ module.exports = function processRequest(options, request, response) {
...
                 if (shouldWriteHead) {
                     shouldWriteHead = false;
+                    // Preload the loader script when atleast
+                    // one fragment is present on the page
+                    if (isFragmentFound) {
+                        const loaderScript = getLoaderScript(
+                            amdLoaderUrl,
+                            request.headers
+                        );
+                        loaderScript !== '' &&
```
mo-gr: any reason not to use explicit `if () {}` here?

lib/utils.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const LinkHeader = require('http-link-header');
+
+const getCrossOrigin = (url = '', host) => {
+    if (host && url.indexOf(`://${host}`) < 0) {
```
mo-gr: `url.includes(...)` is a tiny bit more expressive

lib/utils.js
mo-gr: Then maybe replace the `for now` with `until digest spec is implemented by most browsers` to help future maintainers

vigneshshanmugam: Will do.

lib/request-handler.js
vigneshshanmugam: To avoid the typecasting.

lib/utils.js
vigneshshanmugam: done

mo-gr:
:+1:

vigneshshanmugam:
👍

PR #205, opened 2017-12-18
(fix) Pipe the loader(AMD) script url from extended options

vigneshshanmugam:
+ Missed to while doing preloading feature, was missed easily because of no tests. The preloaded script url was different from piped script url
+ Added the tests for piping.

### DIFF
--- a/tests/tailor.js
+++ b/tests/tailor.js
@@ -4,9 +4,12 @@ const http = require('http');
 const nock = require('nock');
 const sinon = require('sinon');
 const zlib = require('zlib');
+const { readFileSync } = require('fs');
+const { resolve } = require('path');
 const { TEMPLATE_NOT_FOUND } = require('../lib/fetch-template');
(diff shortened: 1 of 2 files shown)


tests/tailor.js
```diff
@@ -203,6 +216,40 @@ describe('Tailor', () => {
...
+            mockTemplate.reset();
+            withPipe.close(done);
+        });
+
+        it('should stream pipe definition with loader in the head', done => {
+            getResponse('http://localhost:8083/test')
+                .then(response => {
+                    assert.equal(
+                        response.headers.link,
+                        '<https://loader>; rel="preload"; as="script"; nopush; crossorigin'
```
vigneshshanmugam: checks both preloading and piping has same amd loader url

mo-gr:
:+1:

vigneshshanmugam:
👍

PR #216, opened 2018-02-07
Change hooks to account for multiple script assets

vigneshshanmugam:
+ Perf script handles the case of multiple script assets and measures the overall impact of scripts
+ Uses [puppeteer](https://github.com/GoogleChrome/puppeteer/) for testing the metrics captured from performance entries
+ Made some changes for introducing artificial delays in scripts to measure things close to real world scenario
+ Full blown test including how and when performance timings would happen.

I explained all the details on how the mark and measure events would happen. It would be better to have a look with this current timeline to understand it in a better way. 
screen shot 2018-02-08 at 14 51 16

## Checking product & footer in detail

screen shot 2018-02-08 at 14 52 46

+ Timeline of footer includes the start of the single script till all the scripts from the fragments are done executing including the fetching time of script tags. 
+ Even if they are executed in parallel which would happen for fragments that are lazily rendered, The timeline would work like a charm


vigneshshanmugam:
Figured out a corner case when the scripts are lazy rendered using promise and they can interleave while executing. Will change the script to account for that as well.

examples/fragment-performance/hooks.js
```diff
@@ -0,0 +1,188 @@
...
+     * We should not have any marks for headers script tags
+     *
+     * 0 -> denoted script 1
+     */
+
+    const headerS1Mark = get(mark, 'header0', 'startTime');
+    assert.equal(headerS1Mark, undefined, 'Header s1 mark should be undefined');
+
+    /**
+     * Footer has 2 scripts
```
DeTeam: 3?

DeTeam:
👍

vigneshshanmugam:
👍

DeTeam:
:+1:

PR #225, opened 2018-04-06
account for lazy streamed fragment scripts

vigneshshanmugam:
While measuring timing groups, we forgot to account for fragments that might be streamed asynchronously or streamed lazily on the page. When this happens, the Pipe start would fire a lot late and script execution would not even fire which results in wrong metrics being captured

I have changed the logic to keep adding timing information to the same group name whenever we find fragments of same timing group. This would ensure that even if have fragments that are streamed later, we get the correct value. 

We can use Tailor custom performance API, `addPerfEntry` to add all the timings that we have captured.

** Downside of the approach is that we wont see the timing groups information in Chrome's User timing view **


DeTeam:
:+1:

vigneshshanmugam:
👍

PR #227, opened 2018-04-18
fix: add null check while lazy rendering

vigneshshanmugam:
+ if fragment init code returns null, pipe would explode and throw runtime error
+ Thanks to Ingvar Stepanyan - https://twitter.com/RReverser/status/986687943602352128


DeTeam:
:+1:

ruiaraujo:
👍

Issue #231, opened 2018-05-22
Add tracing instrumentation


lmineiro:
Tailor is often used as part of large-scale distributed systems and providing out of the box tracing support will extend the observability already present.

[Opentracing](https://opentracing.io) has gained enough traction and community support to be considered the current standard way of providing distributed tracing across polyglot systems.

The goal of this issue is to [instrument the basic operations first](https://github.com/opentracing/opentracing.io/blob/master/_docs/pages/instrumentation/instrumenting-large-systems.md#crawl-walk-run), preparing the environment to provide a global tracer which can later (with separate issues) be used to instrument all other missing operations.

vigneshshanmugam:
closing in favor of #232 #236

PR #232, opened 2018-05-24
added opentracing instrumentation

lmineiro:
This addresses #231 

The server request handling and fetching of fragments are instrumented. There are some caveats with the instrumentations of Fragment.fetch():
- A new optional argument is passed to the nested functions so that it's possible to reference the parent Span
- It's possible to pass a custom fetchFragment function which doesn't handle the new parentSpan parameter. To guarantee span context propagation, those custom functions should handle the new parameter and propagate it using the appropriate carrier. For HTTP the default implementation can be used as a reference

Tailor will use, by default the "NoOpTracer" which is part of the standard OpenTracing libraries. This tracer implementation discards all the spans collected.

To use a Tracer from other vendors (Jaeger, Lighstep, etc) provide the Tracer instance to the new options attribute of Tailor `tracer: yourCustomerTracer`.

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -95,14 +95,16 @@ module.exports = class Fragment extends EventEmitter {
             : this.attributes.url;
 
         const spanOptions = parentSpan ? { childOf: parentSpan } : {};
-        const span = tracer.startSpan('fetch_fragment', spanOptions);
-        span.setTag('fragment-url', url);
-        span.setTag('fragment-id', this.attributes.id || 'unnamed');
-        span.setTag('fallback', isFallback);
-        span.setTag('primary', this.attributes.primary || false);
-        span.setTag('async', this.attributes.async || false);
-        span.setTag('public', this.attributes.public || false);
-        span.setTag(Tags.SPAN_KIND, Tags.SPAN_KIND_RPC_CLIENT);
+        const span = tracer.startSpan('fetch-fragment', spanOptions);
+        span.addTags({
+            'fragment-url': url,
+            'fragment-id': this.attributes.id || 'unnamed',
+            fallback: isFallback,
+            primary: this.attributes.primary,
+            async: this.attributes.async,
+            public: this.attributes.public,
+            [Tags.SPAN_KIND]: Tags.SPAN_KIND_RPC_CLIENT
+        });
         this.requestFragment(url, this.attributes, request, span)
             .then(
                 res => this.onResponse(res, isFallback),
@@ -127,9 +129,7 @@ module.exports = class Fragment extends EventEmitter {
                     }
                 }
             )
-            .then(() => {
-                span.finish();
-            });
+            .then(() => span.finish());
         // Async fragments are piped later on the page
         if (this.attributes.async) {
             return Buffer.from(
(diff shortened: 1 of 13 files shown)


vigneshshanmugam:
@lmineiro the yarn.lock file is having lot of duplicates, are you using older version of node? Ideally we should see only opentracing added as part of yarn lock file.

lib/request-handler.js
```diff
@@ -29,6 +32,15 @@ const FRAGMENT_EVENTS = [
  * @param {Object} response - HTTP response stream of middleware
  */
 module.exports = function processRequest(options, request, response) {
+    const parentSpanContext = tracer.extract(
```
vigneshshanmugam: tracer should be present after this.emit("start"), otherwise our overhead wont calculate the time spent in starting tracer context.

lib/request-handler.js
```diff
@@ -64,12 +76,18 @@ module.exports = function processRequest(options, request, response) {
 
     const handleError = err => {
         this.emit('error', request, err);
+        span.setTag(Tags.ERROR, true);
+        span.log({
+            message: err.presentable,
```
vigneshshanmugam: the presentable info can sometimes be a default template, should we log this?

lib/fragment.js
```diff
@@ -80,33 +83,53 @@ module.exports = class Fragment extends EventEmitter {
...
-            res => this.onResponse(res, isFallback),
-            err => {
-                if (!isFallback) {
-                    const { fallbackUrl } = this.attributes;
-                    if (fallbackUrl) {
-                        this.emit('fallback', err);
-                        this.fetch(request, true);
+        const spanOptions = parentSpan ? { childOf: parentSpan } : {};
+        const span = tracer.startSpan('fetch_fragment', spanOptions);
+        span.setTag('fragment-url', url);
```
vigneshshanmugam: can we use addTags instead of setTag here?

lib/fragment.js
```diff
@@ -80,33 +83,53 @@ module.exports = class Fragment extends EventEmitter {
...
+        span.setTag('fragment-id', this.attributes.id || 'unnamed');
+        span.setTag('fallback', isFallback);
+        span.setTag('primary', this.attributes.primary || false);
```
vigneshshanmugam: we dont need || false here, we already assign default values in the constructor 
https://github.com/zalando/tailor/blob/477c65166f2c09e3c6f5928ff87b4047b5dc6215/lib/fragment.js#L61-L69

lib/request-handler.js
lmineiro: I actually, deliberately, put it before so that the Span duration would itself contain the overhead :)
We may be having some redundant measurements for some time

lib/request-handler.js
lmineiro: Most probably not. I don't have enough knowledge about those details. Please suggest what you believe is more reasonable

lib/fragment.js
lmineiro: I don't think the opentracing API spec has a addTags. This is also a deliberate attempt at using the spec as is to also understand how much such approach pollutes the codebase. We could export already a helper function in the tracing.js for that purpose. 
Something like:
```javascript
addTags(span, {...});
```
Would that work for you?

lib/fragment.js
lmineiro: Cool. Will remove.

vigneshshanmugam:
👍

vigneshshanmugam:
👍

lmineiro:
:+1:

PR #235, opened 2018-05-28
fix: handle parsing & primary error properly

vigneshshanmugam:
+ If we respond with a empty message with Content-Type: 'text/html', the browser will throw error stating as **This page isn’t working** losing the original error page. 
+ Now we respond with 500 status code and Internal Server Error as the response when there is a parse error or when there is no fallback provided for primary fragment. 

This is a very old bug and it was introduced when we started supporting fallbacks and I noticed after we started the Opentracing instrumentation, so Thanks to @lmineiro  😄

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -23,6 +23,8 @@ const FRAGMENT_EVENTS = [
     'fallback',
     'warn'
 ];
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
👍

DeTeam:
:+1:

PR #238, opened 2018-06-01
renamed operation name

lmineiro:
After giving it some more thought, I believe the most appropriate name for the edge request in Tailor should be `compose-page`.

It follows the rationale that Tailor is a **Composition Service** and the requests it receives are requests aimed at **composing a page**. `compose-page` follows [the conventions](https://github.com/opentracing/specification/blob/master/specification.md#start-a-new-span) that we also have adopted so far - a verb and the object related to such verb.


aryszka:
:+1:

vigneshshanmugam:
👍

PR #243, opened 2018-07-03
introduce multiple subscribers for on done hook

vigneshshanmugam:
+ Having multiple subscribers in onDone performance hook will enable us do post processing of all the metrics and also do reporting in different parts of the application.
+ We do not need multiple subscribers in other hooks since it will make it hard to measure the fragment script exec times. 
+ We are not using pub,sub pattern where we would normally push in to array and call the subscribers, Instead we are chaining the callbacks to make the code simpler and also backwards compatible.

### DIFF
--- a/examples/fragment-performance/hooks.js
+++ b/examples/fragment-performance/hooks.js
@@ -113,7 +113,7 @@ async function analyseHooks(mark, measure, entries) {
     /**
      * Performance Measures
      * 3 - timing groups - abovethefold, belowthefold, interactive
-     * 1 - all done
+     * 2 - all done
      * 1- primary fragemnt done
      * 3 including all the fragments
      *  -> Header - 1
@@ -124,7 +124,7 @@ async function analyseHooks(mark, measure, entries) {
      *
      * 3 script tags in Footer fragment
      */
-    assert.equal(measure.length, 10, 'No of measure entries must be 10');
+    assert.equal(measure.length, 11, 'No of measure entries must be 10');
 
     /**
      * No of tailor entries must be three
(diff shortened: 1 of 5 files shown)


examples/fragment-performance/hooks.js
```diff
@@ -113,7 +113,7 @@ async function analyseHooks(mark, measure, entries) {
     /**
      * Performance Measures
      * 3 - timing groups - abovethefold, belowthefold, interactive
-     * 1 - all done
+     * 2 - all done
```
aryszka: what is the meaning of these numbers?

src/pipe.js
```diff
@@ -205,6 +205,18 @@
...
     }
 
+    /**
+     * Preserve and execute the functions in the order
+     * the callbacks were registered
+     */
+    function onDoneHook(cb) {
+        var prevCb = hooks['onDone'];
+        hooks['onDone'] = function() {
+            prevCb();
```
DeTeam: try-catch?

vigneshshanmugam: It should blew up. Its better to not measure anything and throw instead of handling error and moving on with wrong measurements.

DeTeam: I'd assume `done` event to be more generic to be honest as a consumer. Using it for perf measurements (what we do) is only one use-case (though, a real one!).

Why I asked, cause with normal `events` in node that'd be the case — chain doesn't break when there's an error in one of the handlers.

DeTeam:
:+1:

src/pipe.js
DeTeam: Mb adding one more sentence https://github.com/zalando/tailor/blob/master/docs/hooks.md here would resolve potential issues?

vigneshshanmugam: Done.

DeTeam:
:+1:

vigneshshanmugam:
👍

PR #247, opened 2018-08-10
(perf): minor optimization on attributes extraction

vigneshshanmugam:
small optimisation on fragment attributes computation
+ attributes object shape is fixed to minimal set of keys
+ pipe attributes computation is done once instead of multiple times. 

### Before
```
Stat         Avg     Stdev  Max
Latency (ms) 435.94  107.18 984.88
Req/Sec      224.9   47.2   278
Bytes/Sec    2.93 MB 621 kB 3.64 MB
```

### After
```
Stat         Avg     Stdev  Max
Latency (ms) 382.2   93.89  994.23
Req/Sec      257     57.96  315
Bytes/Sec    3.36 MB 768 kB 4.13 MB
```

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -72,10 +72,9 @@ module.exports = class Fragment extends EventEmitter {
         this.attributes = getAttributes(tag, context);
         this.index = index;
         this.maxAssetLinks = maxAssetLinks;
-        this.getPipeAttributes = () =>
-            pipeAttributes(
-                Object.assign({}, { id: this.index }, tag.attributes)
-            );
+        this.pipeAttributes = pipeAttributes(
+            Object.assign({}, { id: this.index }, tag.attributes)
+        );
         this.requestFragment = requestFragment;
         this.pipeInstanceName = pipeInstanceName;
         this.stream = new PassThrough();
@@ -244,7 +243,7 @@ module.exports = class Fragment extends EventEmitter {
 
         const range = [this.index, this.index + this.scriptRefs.length - 1];
         const fragmentId = id || range[0];
-        const attributes = Object.assign({}, this.getPipeAttributes(), {
+        const attributes = Object.assign({}, this.pipeAttributes, {
             id: fragmentId,
             range
         });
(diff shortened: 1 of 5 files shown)


vigneshshanmugam:
👍

lib/fragment.js
```diff
@@ -24,12 +31,17 @@ const getAttributes = (tag, context) => {
         Object.assign(attributes, fragmentCtxt);
     }
 
-    return Object.assign({}, attributes, {
-        url: attributes.src,
+    const { src, async, primary, public: isPublic, timeout } = attributes;
```
mo-gr: why `public: isPublic`

lib/fragment.js
```diff
@@ -9,6 +9,13 @@ const zlib = require('zlib');
 const { globalTracer, Tags } = require('opentracing');
 const tracer = globalTracer();
 
+const getValue = value => {
```
mo-gr: this function seems misnamed. It's not getting the value, it's doing a kind of extended truthiness-check, isn't it? Maybe better named `hasValue`?

lib/fragment.js
vigneshshanmugam: reserved keyword, changing for async as well.

vigneshshanmugam:
@mo-gr Need Github PR approval.

lib/fragment.js
```diff
@@ -57,23 +75,12 @@ module.exports = class Fragment extends EventEmitter {
...
-
+        this.attributes = getAttributes(tag, context);
         this.index = index;
         this.maxAssetLinks = maxAssetLinks;
-        this.getPipeAttributes = () =>
-            pipeAttributes(
-                Object.assign({}, { id: this.index }, tag.attributes)
-            );
+        this.pipeAttributes = pipeAttributes(
+            Object.assign({}, { id: this.index }, tag.attributes)
```
ruiaraujo: nit: the empty shape is pointless.

lib/fragment.js
```diff
@@ -96,22 +103,29 @@ module.exports = class Fragment extends EventEmitter {
...
             [Tags.SPAN_KIND]: Tags.SPAN_KIND_RPC_CLIENT,
             [Tags.HTTP_URL]: url,
-            id: this.attributes.id || 'unnamed',
             fallback: isFallback,
-            primary: this.attributes.primary,
-            async: this.attributes.async,
-            public: this.attributes.public
+            public: isPublic,
+            async: isAsync,
+            id,
```
ruiaraujo: The behaviour for id has changed, default arguments only apply to undefined values. Null or empty string will passthrough.

vigneshshanmugam: good catch. changing it back

vigneshshanmugam:
@ruiaraujo Done

vigneshshanmugam:
👍

ruiaraujo:
👍

PR #248, opened 2018-08-12
(perf): improve parsing of link header from fragments

vigneshshanmugam:
+ Roll our own fine tuned parse link header implementation based on https://github.com/thlorenz/parse-link-header
+ Avoid double parsing of link headers on fragment response and while preloading
+ Keep fragment object shape static

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -5,7 +5,6 @@ const PassThrough = require('stream').PassThrough;
 const zlib = require('zlib');
 const ContentLengthStream = require('./streams/content-length-stream');
 const parseLinkHeader = require('./parse-link-header');
-const { getFragmentAssetUris } = require('./utils');
 
 const { globalTracer, Tags } = require('opentracing');
 const tracer = globalTracer();
@@ -15,6 +14,20 @@ const hasValue = value => {
         return true;
     }
     return false;
+};
+
+const getFragmentAssetUris = (refs, assetSize) => {
+    const scriptUris = [];
+    const styleUris = [];
+
+    for (const ref of refs) {
+        if (ref.rel === 'fragment-script') {
+            scriptUris.push(ref.uri);
+        } else if (ref.rel === 'stylesheet') {
+            styleUris.push(ref.uri);
+        }
+    }
+    return [scriptUris.slice(0, assetSize), styleUris.slice(0, assetSize)];
 };
 
 /**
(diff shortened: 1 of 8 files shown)


lib/utils.js
```diff
@@ -80,6 +68,20 @@ const getFragmentAssetsToPreload = (headers, { host } = {}) => {
     return assetsToPreload;
 };
 
+function getFragmentAssetUris(refs, assetSize) {
```
ruiaraujo: Why put this in utils, is it used anywhere else?

Are we planning to add this to the exported API?

tests/parse-link-header.js
```diff
@@ -0,0 +1,49 @@
...
+        ]);
+    });
+
+    it('filters invalid header links', () => {
+        const linkHeader = 'http://a.com/app.js; rel="script"';
+
+        assert.deepStrictEqual(parseLinkHeader(linkHeader), []);
+    });
+
+    it('filters invalid rel attributes', () => {
```
ruiaraujo: What is an invalid rel attr?
Empty? Can I use in rel="styleshit"?

lib/fragment.js
```diff
@@ -16,6 +16,20 @@ const hasValue = value => {
     return false;
 };
 
+const getFragmentAssetUris = (refs, assetSize) => {
+    const scriptUris = [];
+    const styleUris = [];
+
+    for (const ref of refs) {
+        if (ref.rel === 'fragment-script') {
```
ruiaraujo: Do we have tests for this filtering behaviour?

tests/parse-link-header.js
vigneshshanmugam: Ofcourse you can lol. its valid. But wont be loaded because of the check in getFragmentAssetUris.

lib/fragment.js
vigneshshanmugam: Yes, tests/tailor.js.

tests/parse-link-header.js
vigneshshanmugam: But the test if for rel="" and no rel attribute at all. do you have any better naming suggestion if its confusing?

ruiaraujo:
👍

tests/parse-link-header.js
ruiaraujo: maybe it could be rephrased... but it is fine.

vigneshshanmugam:
👍

PR #262, opened 2018-09-25
mark span as error on unzip failure

vigneshshanmugam:
+ when error happens during unzipping, its a broken response and it should be marked as error.


vigneshshanmugam:
👍

ruiaraujo:
:+1:

DeTeam:
:+1:

PR #270, opened 2018-11-14
opentracing: add timeout tag to fragment spans

vigneshshanmugam:
+ Helps in tracing when there are timeout issues with the remote fragment endpoint.

### DIFF
--- a/lib/fragment.js
+++ b/lib/fragment.js
@@ -124,7 +124,8 @@ module.exports = class Fragment extends EventEmitter {
             id,
             primary,
             async: isAsync,
-            public: isPublic
+            public: isPublic,
+            timeout
         } = this.attributes;
 
         span.addTags({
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
👍

mo-gr:
:+1:

lmineiro:
This looked more appropriate as a log field if it will contain just the info about the current fragment timeout. It's expected to change and not very deterministic anyway. As information, it would be a good log field, imho

vigneshshanmugam:
I agree with you, But I had put it as tag for two reasons
1. Need to keep the consistency of existing attributes and do not treat this field in any different way. 
2. Since logs fields aren't searchable and for debugging purposes once I can get the timeout information, I can still search for timeout even though it isn't deterministic. 

We already log on z-tailor with the timeout error with specified read and connection timeouts which maps with the error.

Issue #274, opened 2018-12-05
4.0.0 Release


vigneshshanmugam:
The next major release 4.0.0 of tailor should bring contain these below features. These features were used inside at Zalando and its much more performant than the open source version. So its to upstream those

+ [ ] Should use Keep alive agents (https://github.com/node-modules/agentkeepalive)
+ [ ] Caching of fragments either by using the Id and URL or through cache-control headers
+ [ ] Logging and Metrics(Fragment TTFB, TTLB) that are based on tailor event system
+ [ ] System level metrics like IO/CPU which can be made available via child process 
+ [ ] Template caching on bootup ( Super easy for File system, Probably for now we can skip if its a remote endpoint and have to take care of update mechanism to be generic)
+ [ ] Circuit breakers pool for each fragment
+ [ ] Agent pooling (https://github.com/vigneshshanmugam/agent-pool/)

vigneshshanmugam:
If anyone wants to pick it up specific taks, Let me know.. I can give more context...

vigneshshanmugam:
@franz-josef-kaiser Sure, Please provide ideas on the issue on what you want the behaviour to be when the obligatory ones fail and whether its applicable only for primary or all the fragments.

tobsch:
any progress on this?

vigneshshanmugam:
Was bit busy with other things, I will come back to working on this in few days.

soofka:
Hello @vigneshshanmugam, I've talked to you on Twitter a few days back, we've agreed that I'll create an issue and we'll discuss details here. However, I don't want to duplicate existing issues, and pretty much all of issues or lack of features we've experienced are listed here. So I'm commenting here, let me know if you still prefer separate thread for our support.

Features we are craving for in Tailor:
- keep-alive connection for fragments
- caching of fragments
- circuit breaker for fragments

All of those are mentioned here. Is it still the plan? Are they going to be implemented? Do you need any help with that? If yes, how can we help? :-)

Additionally, there is this point on your list here, "Logging and Metrics(Fragment TTFB, TTLB) that are based on tailor event system". Is it somehow related to hooking system? Does it mean that you are aware of the fact that events described in documentation of hooks (https://github.com/zalando/tailor/blob/master/docs/hooks.md) are never triggered? At least that's what we've experienced. If this is not the same thing, I'll open separate issue with detailed description for this particular bug.

Thanks a lot! :-)

vigneshshanmugam:
[...]

PR #282, opened 2019-01-28
feat: use keep alive agents for all fragment requests

vigneshshanmugam:
+ part of https://github.com/zalando/tailor/issues/274
+ Merge #283 and rerun this test.

### DIFF
--- a/lib/agent-store.js
+++ b/lib/agent-store.js
@@ -1,0 +1,20 @@
+'use strict';
+const Agent = require('agentkeepalive');
+const { SOCKET_TTL } = require('./constants');
+
+class AgentStore {
+    constructor() {
+        const defaultAgentOpts = { socketActiveTTL: SOCKET_TTL };
+        this.httpAgent = new Agent(defaultAgentOpts);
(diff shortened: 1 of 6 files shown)


lib/agent-store.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+const Agent = require('agentkeepalive');
+const { SOCKET_TTL } = require('./constants');
+
+class AgentStore {
+    constructor() {
+        const defaultAgentOpts = { socketActiveTTL: SOCKET_TTL };
```
ruiaraujo: The socket ttl should be configurable

vigneshshanmugam: Hmmm, I was thinking the same as well, Plus have to come up with custom config options for circuit breaker pool settings + metrics + caching of fragments.. 

I will come up with some proposal.

vigneshshanmugam: Thought for a while about this and I think we can customize all of the logic in a client library like perron(or create one specific for tailor - or give PR to perron to make it work). Perron already supports circuit breaker logic for clients and we can easy add cache specific logic as well..

I will give PR to perron once i have sometime to work on some features.

PR #297, opened 2019-04-01
fix: requestFragment option type

armand1m:
# Problem

Currently, when trying to create a `requestFragment` handler in typescript, the current type says that I should implement the fragment handler like this:

```ts
import { IncomingMessage } from "http";
import { Url } from "url";

import requestFragmentFn from "node-tailor/lib/request-fragment";

/** had to hardcode these from the index.d.ts since they're not exported */
nterface IAttributes {
  id: string;
  src: string;
  async?: boolean;
  fallbackUrl?: string;
  primary?: boolean;
  public?: boolean;
  [key: string]: any;
}

const handleFragment = (
  /** the types are currently demanding that I should define this filterHeaders */
  filterHeaders: (attributes: IAttributes, req: IncomingMessage) => object,
  fragmentUrl: Url,
  fragmentAttributes: IAttributes,
  request: IncomingMessage,
  span = null,
) => {
  /* imagine some logic here */

  return requestFragmentFn(filterHeaders)(fragmentUrl, fragmentAttributes, request, span);
};

export default handleFragment;
```

```ts
import Tailor from "node-tailor";
import requestFragment from "./handlers/fragment";

const tailor = new Tailor({
  requestFragment,
});
```

which results in problems since the actual call in the source code [is not actually like this](https://github.com/zalando/tailor/blob/master/lib/fragment.js#L142)

This PR fixes it by making it respect the actual implementation usage.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,7 +39,7 @@ declare class Tailor extends EventEmitter {
     , maxAssetLinks?: number
     , pipeAttributes?: (attributes: Attributes) => object
     , pipeInstanceName?: string
-    , requestFragment?: (filterHeaders: (attributes: Attributes, req: IncomingMessage) => object, url: Url, attributes: Attributes, req: IncomingMessage, span?: Span) => Promise<ServerResponse>
+    , requestFragment?: (url: Url, attributes: Attributes, req: IncomingMessage, span?: Span) => Promise<ServerResponse>
     , templatesPath?: string
     , tracer?: Tracer
   })


vigneshshanmugam:
👍

armand1m:
I'll also have to open one for the `fetchTemplate` soon as the types are also misleading on that one

vigneshshanmugam:
@armand1m Sure, Please do

vigneshshanmugam:
👍

oporkka:
Is this PR waiting for something to be added?

armand1m:
> Is this PR waiting for something to be added?

CI is failing with messages from saucelabs issues I cannot solve:

```
  1) Frontend test should open the page and initialise three fragments, each requiring two scripts in internet explorer:
     Error: [sauceJobStatus(false)] Missing auth token.
      at PromiseWebdriver.Webdriver._sauceJobUpdate (node_modules/wd/lib/webdriver.js:186:17)
      at PromiseWebdriver.commands.sauceJobStatus (node_modules/wd/lib/commands.js:2324:24)
```

atkinchris:
I think `requestFragment` needs the initial `filterHeaders` function. If it's passed in `options`, it won't be called with `filterHeaders` - it'll just be assigned over it.

https://github.com/zalando/tailor/blob/571b6d545e99957705c507cdba09a6c0c2871985/index.js#L65-L81

If having the implementing application close over `filterHeaders` is not desired, should there be an additional line of:

```ts
options.requestFragment = options.requestFragment(filterRequestHeaders)
```

PR #298, opened 2019-04-01
fix: fetchTemplate option type

armand1m:
# Problem

Currently, trying to implement  your own fetchTemplate in typescript respecting the implementation will throw this type error:

```
Type '(request: IncomingMessage, parseTemplate: any) => any' is not assignable to type '(templatesPath: string, baseTemplateFn: (path: string) => string) => Promise<any>'.
```

This pull request fixes the types according to the current implementation, also adding a `ParseTemplateFunction` type


vigneshshanmugam:
👍

addityasingh:
👍

PR #301, opened 2019-06-19
Fix README link to basic example with CSS+JS Link headers

robinwhittleton:
The existing `example/fragment.js` doesn’t exist; I’m guessing this is the correct example to link to if showing off how to use the `Link` header is important here


DeTeam:
👍

vigneshshanmugam:
👍

addityasingh:
👍

PR #302, opened 2019-06-19
feat: Run puppeteer test for PR and let the selenium fail for PR

### LINKED ISSUE #300: Frontend tests fail for external contributors due to Saucelabs
Frontend tests fail for external contributors due to Saucelabs. Examples:
- https://travis-ci.org/zalando/tailor/jobs/468800950
- https://travis-ci.org/zalando/tailor/jobs/514400191
- https://travis-ci.org/zalando/tailor/jobs/513914175

In order to resolve this, we need to either check if we really need saucelabs or using something like puppetteer would be just fine

addityasingh:
Close #300 
- Run puppeteer test for PR and let the selenium fail for PR 
- Change Travis matrix to jobs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,41 @@
 sudo: false
 language: node_js
-node_js:
-- '8'
-- '10'
-- '11'
 cache:
   directories:
   - node_modules
 before_script:
 - yarn run lint
-script:
-- yarn run coverage
-- if [ -n "${FRONTEND-}" ]; then yarn run test:frontend ; fi
-- yarn run test:frontend:hooks
 after_script:
 - yarn run codecov
+stages:
+  - "Test"
+  - "Test frontend"
+    if: branch = master
 matrix:
   fast_finish: true
-  allow_failures:
-    - script: yarn run test:frontend
   include:
+    - stage: "Test"
+      - node_js: 8
+        script:
+        - yarn run coverage
+        - yarn run test:frontend:hooks
+      - node_js: 10
+        script:
+        - yarn run coverage
+        - yarn run test:frontend:hooks
+      - node_js: 11
+        script:
+        - yarn run coverage
+        - yarn run test:frontend:hooks
+      - node_js: node
+        script:
+        - yarn run coverage
+        - yarn run test:frontend:hooks
+    - stage: "Test frontend"
+      - node_js: node
+        script:
+        - yarn run test:frontend
+  allow_failures:
     node_js: node
     env: FRONTEND=true
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,14 @@ before_script:
 - yarn run lint
 script:
 - yarn run coverage
-- if [ -n "${FRONTEND-}" ]; then yarn run test-frontend:cross-browser ; fi
-- yarn run test-frontend
+- if [ -n "${FRONTEND-}" ]; then yarn run test:frontend ; fi
+- yarn run test:frontend:hooks
 after_script:
 - yarn run codecov
 matrix:
   fast_finish: true
   allow_failures:
-    - script: yarn run test-frontend:cross-browser
+    - script: yarn run test:frontend
   include:
     node_js: node
     env: FRONTEND=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,6 @@ language: node_js
 cache:
   directories:
   - node_modules
-before_script:
-- yarn run lint
-after_script:
-- yarn run codecov
 stages:
   - lint
   - test:frontend:hooks
@@ -18,6 +14,7 @@ jobs:
     - script: yarn run test:frontend
   include:
     - stage: lint
+      node_js: 11
       script: yarn run lint
     - stage: test:frontend:hooks
       node_js: 8
@@ -32,6 +29,7 @@ jobs:
       node_js: node
       script: yarn run test:frontend
     - stage: coverage
+      node_js: 11
       script: yarn run coverage
 addons:
   sauce_connect:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: node_js
 cache:
   directories:
   - node_modules
+after_script:
+- yarn run codecov
 stages:
   - lint
   - test
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,36 +8,31 @@ before_script:
 after_script:
 - yarn run codecov
 stages:
-  - "Test"
-  - "Test frontend"
-    if: branch = master
-matrix:
+  - lint
+  - test:frontend:hooks
+  - test:frontend
+  - coverage
+jobs:
   fast_finish: true
-  include:
-    - stage: "Test"
-      - node_js: 8
-        script:
-        - yarn run coverage
-        - yarn run test:frontend:hooks
-      - node_js: 10
-        script:
-        - yarn run coverage
-        - yarn run test:frontend:hooks
-      - node_js: 11
-        script:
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "scripts": {
     "test": "mocha --harmony tests/**",
-    "test-frontend": "mocha --harmony examples/multiple-fragments-with-custom-amd/test.js",
+    "test-frontend": "mocha --harmony examples/fragments-performance/hooks.js",
```
vigneshshanmugam: need to be `node examples/fragments-performance/hooks.js` 

Also rename it to `test:frontend:hooks`

addityasingh: I have created a new file `examples/fragments-performance/test.js` instead and will run it via `mocha` to have a test spec instead of just running a node script

.travis.yml
```diff
@@ -1,24 +1,40 @@
...
+    if: branch = master
 matrix:
   fast_finish: true
   include:
-    node_js: node
-    env: FRONTEND=true
+    - stage: "Test"
+      - node_js: 8
+      - script:
+        - yarn run coverage
```
vigneshshanmugam: Lets move coverage as separate step. No need to run it on every version. So the stage can be combined like this. 

stages:
  - lint
  - test (run multiple node version(8,10,12) since it will run all unit tests)
  - test:frontend
  - coverage

With the above stages, you can do matrix now.

examples/fragment-performance/test.js
```diff
@@ -0,0 +1,86 @@
...
+            maxAssetLinks: 3
+        });
+        server = http.createServer((req, res) => {
+            if (req.url === '/favicon.ico') {
+                res.writeHead(200, { 'Content-Type': 'image/x-icon' });
+                return res.end('');
+            }
+            return tailor.requestHandler(req, res);
+        });
+        server.listen(8080);
```
vigneshshanmugam: test should start only after server is started, 

before(done => server.listen(0, done));

examples/fragment-performance/test.js
```diff
@@ -0,0 +1,86 @@
...
+    
+        const fragment3 = http.createServer(
+            serveFragment('Footer', 'http://localhost:8083', 3, true)
+        );
+        fragment3.listen(8083);
+    })
+
+    after(() => {
+        if(server != null) {
+            server.close();
```
vigneshshanmugam: same here, server.close(done)

examples/fragment-performance/test.js
```diff
@@ -0,0 +1,86 @@
...
+                JSON.parse(metrics[1]),
+                JSON.parse(metrics[2])
+            ];
+            await analyseHooks(mark, measure, entries);
+            await browser.close();
+        } catch (e) {
+            console.error(e);
+            process.exit(1);
+        }
+    }).timeout(8000); // Add increased timeout since the test runs inside puppeteer
```
vigneshshanmugam: use timeout from mocha CLI flags and use `done` function since its an async test.

.travis.yml
addityasingh: I am still trying the combinations but the travis matrix is really poorly documented for the use case and `allow_failures` doesn't work as expected. Still figuring this out

examples/fragment-performance/test.js
addityasingh: We don't need the `done` for an async test if the assertion is done at the right time, which in this case is calling the `analyseHooks` method call and the control is at the end of this function. But since we are not calling any assertion directly in this test, I will add the done for making it explicit

.travis.yml
vigneshshanmugam: Let me give you an example

https://github.com/elastic/apm-agent-rum-js/blob/b57be6f7052233fc421f8699ad185d8e367ca57a/.travis.yml#L21-L61

examples/fragment-performance/test.js
vigneshshanmugam: The reason is suggested for done is, if we failed to capture any promise rejections and done is never called and we wont be able to see any meaningful error message from mocha.

addityasingh: Added

.travis.yml
```diff
@@ -1,24 +1,36 @@
...
-- yarn run lint
-script:
-- yarn run coverage
-- if [ -n "${FRONTEND-}" ]; then yarn run test-frontend ; fi
-after_script:
-- yarn run codecov
-matrix:
+stages:
+  - lint
+  - test:frontend:hooks
```
vigneshshanmugam: what about unit tests?  We have to run unit tests on 8, 10 and 12 (remove 11) and also frontend:hooks can run only on any single version, It is not really required since it does not change the functionality.

examples/fragment-performance/test.js
addityasingh: On checking the travis failures, it seems we can only use 1 of `async` or `done`. I couldn't recall earlier why I didn't have but now the error logs show the same: https://travis-ci.org/zalando/tailor/jobs/548746746
> 1) Fragment performance in browser should be able to capture performance metrics in browser:
     Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.
      at Context.it (examples/fragment-performance/test.js:81:13)

.travis.yml
addityasingh: > what about unit tests? We have to run unit tests on 8, 10 and 12 (remove 11) 
Added

> also frontend:hooks can run only on any single version, It is not really required since it does not change the functionality.

I didn't get this one

examples/fragment-performance/test.js
vigneshshanmugam: Hmmm. interesting, let's remove it then.

.travis.yml
vigneshshanmugam: `npm run test` -> unit tests, We don't need to run `npm run test:frontend:hooks` on all versions.

addityasingh: Which one do you recommend? I would keep it to the latest, in this case 12

.travis.yml
```diff
@@ -1,24 +1,46 @@
...
+    - stage: test
+      node_js: 10
+      script: 
+        - yarn run test
+    - stage: test
+      node_js: 12
+      script: 
+        - yarn run test
+        - yarn run test:frontend:hooks
+    - stage: test
```
vigneshshanmugam: I don't think we need this stage.

examples/fragment-performance/test.js
```diff
@@ -0,0 +1,86 @@
+const puppeteer = require('puppeteer');
+const assert = require('assert');
+const http = require('http');
+
+const Tailor = require('../../index');
+const serveFragment = require('../multiple-fragments-with-custom-amd/fragment');
+const { analyseHooks } = require('./hooks');
```
vigneshshanmugam: since the function inside hooks is IIFE, its getting executed twice in this test.https://github.com/zalando/tailor/blob/654b4b83b4809775d8bc5962206d0595c8a417b5/examples/fragment-performance/hooks.js#L4

Can you extract that function and use the same in test as well. 



Check travis logs - https://travis-ci.org/zalando/tailor/jobs/548757347

.travis.yml
addityasingh: I would still suggest that we test on latest node as well. It doesn't have any significant negative impact on the pipeline time, since the job runs in parallel to other test jobs for node 8, 10 and 12

vigneshshanmugam: Sure, not a big deal.

addityasingh:
👍

vigneshshanmugam:
👍

boopathi:
👍

Issue #327, opened 2020-05-08
requestFragment type definition does not match implementation


atkinchris:
The type definition for `requestFragment` is a single function that takes 5 arguments and returns a promise.
https://github.com/zalando/tailor/blob/7283c2fb851d1ca2916dc996805acbef1c958177/index.d.ts#L42

However, the implementation is a function that takes a single argument, and returns a function that takes 4 arguments and returns a promise.
https://github.com/zalando/tailor/blob/571b6d545e99957705c507cdba09a6c0c2871985/lib/request-fragment.js#L26-L32

I think there's a misplaced closing bracket in the type definition, that it should be:

```ts
requestFragment?: (filterHeaders: (attributes: Attributes, req: IncomingMessage) => object) => (url: Url, attributes: Attributes, req: IncomingMessage, span?: Span) => Promise<ServerResponse> 
```

PR #328, opened 2020-05-08
Correct requestFragment type and initialisation

atkinchris:
This corrects the type definition for requestFragment, which does not match it's implementation. This resolves https://github.com/zalando/tailor/issues/327 and supercedes https://github.com/zalando/tailor/pull/297.

This also correct initialisation of custom `requestFragment` functions, which otherwise are not invoked to close over `filterHeaders`.


atkinchris:
The tests are failing on Node 12, but not on earlier versions (on Travis and local). Is this related to https://github.com/zalando/tailor/pull/322?

